PR #2, opened 2020-07-22
Implement basic query functionality

shulijones:
Implement basic functionality to show a query and get a data table based on it. 

This code: 
displays a query with dropdowns on the home page; 
gets the query in JavaScript; 
sends a GET request to the backend with the query information as parameters; 
chooses a hard-coded data table based on the query parameters; 
makes a fetch request to the Census API for the data table;
and passes it back up to JavaScript, to pass it to Olivia's display function.


src/main/webapp/script.js
```diff
@@ -0,0 +1,19 @@
...
+    .then((response) => {
+      if (response.ok) {
+        response.json().then((jsonResponse) => JSON.parse(jsonResponse))
+        .then((data) => {
+          // data is a 2D array, where the first row is a header row and all
+          // subsequent rows are one piece of data (e.g. for a state or county)
+          displayVisualization(data);
+        });
+      } else {
+        console.log('There was an error');
```
anicr7, line 16: Maybe add the error here from the response.

src/main/java/QueryServlet.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.InputStreamReader;
+
+import com.google.gson.Gson;
+
+/** Servlet that handles census queries from users */
+@WebServlet("/query")
+public class QueryServlet extends HttpServlet {
+
+  // Hardcoded initial choices for which lines of the data table are accessed
+  Map<String, String> lineSelector =
```
anicr7, line 23: Maybe a better name for this? lineSelector does not represent this well enough.

src/main/java/QueryServlet.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.google.gson.Gson;
+
+/** Servlet that handles census queries from users */
+@WebServlet("/query")
+public class QueryServlet extends HttpServlet {
+
+  // Hardcoded initial choices for which lines of the data table are accessed
+  Map<String, String> lineSelector = 
+    Stream.of(
```
anicr7, line 24: Optional: Maybe cleaner to use enums instead.

src/main/java/QueryServlet.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String personType = request.getParameter("person-type");
+    String location = request.getParameter("location");
+
+    URL fetchUrl = 
+      new URL(
+        "https://api.census.gov/data/2018/acs/acs1/spp?get=NAME,S0201_" 
+          + lineSelector.get(personType)
```
anicr7, line 39: Perform basic error checking for this.

src/main/java/QueryServlet.java
shulijones: Hmm. I agree that I think that would be better for now, but later I am envisioning that there will be a much larger map (or some other even more complex data structure, perhaps?) that is being used to match parts of queries to data tables. Do you think it's better to use an enum now knowing we will switch to something else later, or to leave it like this?

anicr7: Resolving. If we do end up using just the string maybe we can revisit it.

src/main/webapp/script.js
shulijones: Done

src/main/java/QueryServlet.java
shulijones: Done (a little bit above this row)

src/main/java/QueryServlet.java
shulijones: Changed to queryToDataRowMapper

src/main/webapp/script.js
```diff
@@ -0,0 +1,20 @@
...
+  const location = query.get('location');
+
+  fetch('/query?person-type=' + personType + '&location=' + location)
+    .then((response) => {
+      if (response.ok) {
+        response.json().then((jsonResponse) => JSON.parse(jsonResponse))
+        .then((data) => {
+          // data is a 2D array, where the first row is a header row and all
+          // subsequent rows are one piece of data (e.g. for a state or county)
+          displayVisualization(data);
```
nsri-google, line 13: since you dont have this function now, you could either do a console.log(data).

shulijones: I did have a log originally. But then I saw that Olivia named her function displayVisualization on her branch, so I changed it to this so that they can merge nicely.

PR #3, opened 2020-07-23
Olivia javascript

otobin:
Added visualizations for states and messages for requests with counties.


src/main/webapp/script.js
```diff
@@ -10,11 +15,49 @@ function passQuery() {
...
+
+ // displayVisualization takes in a data array representing the 2D array returned by the 
+ // census API. The function creates the necessary visualization with the census data.
+function displayVisualization(censusDataArray) {
+  if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
+    google.charts.setOnLoadCallback(drawRegionsMap(censusDataArray));
+  } else {
+    // We currently do not have counties implemented
+    document.getElementById('map').innerHTML = "";
+    document.getElementById('more-info').innerHTML = 'We do not have this information yet.'
```
nsri-google, line 35: Something like "We do not support this visualization yet". Also it might be useful to show the raw data in this (there is datatable visualization I think) but you can add that later.

src/main/webapp/script.js
```diff
@@ -10,11 +15,49 @@ function passQuery() {
...
+}
+
+// createDataArray takes in the data array returned by the census API and reformats it 
+// into a data table that can be processed by the visualization API.
+function createDataArray(censusDataArray) {
+  const vizDataArray = [];
+  censusDataArray.forEach((state) => {
+    vizDataArray.push([state[0], state[1]/1000]);
+  });
+  vizDataArray[0][0] = 'State';
```
nsri-google, line 60: Would this not replace the first value you pushed? How about calling vizDataArray.push(['State', 'Population']) before the foreach loop.

shulijones: I actually wrote this line of code, so I can explain. The data that comes from the census has its own header row, but the google charts visualization doesn't recognize its terms. So this replaces that header row with one that the charts api can understand. There might be a better way to write that, though.

nsri-google: This definitely warrant a comment explaining this in the code.

src/main/webapp/script.js
```diff
@@ -10,11 +15,49 @@ function passQuery() {
...
     });
-}
\ No newline at end of file
+}
+
+ // displayVisualization takes in a data array representing the 2D array returned by the 
+ // census API. The function creates the necessary visualization with the census data.
+function displayVisualization(censusDataArray) {
+  if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
+    google.charts.setOnLoadCallback(drawRegionsMap(censusDataArray));
```
anicr7, line 31: Please add some comments on what the expected format censusDataArray is

src/main/webapp/script.js
```diff
@@ -10,11 +15,49 @@ function passQuery() {
...
+  const shortDataArray = createDataArray(censusDataArray);
+  var data = google.visualization.arrayToDataTable(shortDataArray);
+  var options = {
+    'region': 'US',
+    'resolution': 'provinces',
+    'colorAxis': {colors: ['white', 'blue']}
+  };
+  var chart = new google.visualization.GeoChart(document.getElementById('map'));
+  chart.draw(data, options);
+  document.getElementById('more-info').innerHTML = 'These populations are divided by 1000';
```
anicr7, line 50: Populations(In Thousands) or something similar  might be better.

otobin: fixed

src/main/webapp/script.js
otobin: fixed

src/main/webapp/script.js
otobin: fixed

src/main/webapp/script.js
otobin: added "we do not support this visualization."

PR #4, opened 2020-07-23
Caching

kyracq:
Implement private caching. All queries to QueryServlet are cached.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -7,7 +7,7 @@
   <script src="script.js"></script>
 </head>
 
-<body>
+<body onload="regiesterServiceWorker()">
   <main>
     <div id="query">
       <form id="query-form" onsubmit="passQuery();return false;">
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -8,7 +8,7 @@
   <script src="script.js"></script>
 </head>
 
-<body onload="regiesterServiceWorker()">
+<body onload="registerServiceWorker()">
   <main>
     <div id="query">
       <form id="query-form" onsubmit="passQuery(); return false;">
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -11,8 +11,8 @@ function regiesterServiceWorker() {
   }
 }
 google.charts.load('current', {
-  'packages':['geochart'],
-  'mapsApiKey': 'AIzaSyB5cba6r-suEYL-0E_nRQfXDtT4XW0WxbQ'
+  'packages': ['geochart'],
+  'mapsApiKey': 'AIzaSyB5cba6r-suEYL-0E_nRQfXDtT4XW0WxbQ',
 });
 
 function passQuery() {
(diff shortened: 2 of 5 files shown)


src/main/webapp/service-worker.js
```diff
@@ -0,0 +1,56 @@
+const CACHE_VERSION = 1;
```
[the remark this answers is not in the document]
kyracq: Yes, CACHE_VERSION is just part of the name of the cache and is not changed anywhere, and since the only thing we're caching is census data there will only be one cache in CURRENT_CACHES. These lines and the 'activate' event listener don't do anything right now, but I think it's conventional to have if we were working with multiple caches with different versions. I could remove these for right now since we're only using one cache?

src/main/webapp/script.js
```diff
@@ -1,3 +1,15 @@
+function regiesterServiceWorker() {
```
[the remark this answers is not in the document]
kyracq: Added comment

src/main/webapp/index.html
```diff
@@ -8,7 +8,7 @@
   <script src="script.js"></script>
 </head>
 
-<body>
+<body onload="regiesterServiceWorker()">
```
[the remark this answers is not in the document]
kyracq: Thanks, fixed this

src/main/webapp/service-worker.js
```diff
@@ -0,0 +1,56 @@
...
+const CURRENT_CACHES = {
+  censusData: 'census-data-cache-v' + CACHE_VERSION,
+};
+
+
+self.addEventListener('activate', function(event) {
+  // Delete all caches that aren't named in CURRENT_CACHES.
+  // Useful if multiple versioned caches
```
anicr7, line 9: Useful if multiple versioned caches exist

kyracq:
> How do you guarantee that your service worker has been setup before users can send their queries. It is not a big problem per se just that you might miss catching those queries.

Currently it is possible for the user to submit a query before the service worker is registered, but since registration happens onload, this would only be possible for a very small time frame. I can try to find a way to call the register function before the page is loaded.

src/main/webapp/service-worker.js
kyracq: Done.

src/main/webapp/service-worker.js
```diff
@@ -0,0 +1,56 @@
+const CACHE_VERSION = 1;
+const CURRENT_CACHES = {
+  censusData: 'census-data-cache-v' + CACHE_VERSION,
+};
+
+
+self.addEventListener('activate', function(event) {
```
otobin, line 7: I know essentially nothing about caching, so this may be a very dumb thing to comment on but bear with me, what does it mean to activate? What exactly is being activated? Just trying to figure out where this function is being called.

kyracq: Right after the service worker is registered (called in HTML onload) it's automatically installed and then activated, so the 'activate' event basically happens right after it's registered.

otobin: Ahh I got it. Thank you for clarifying that for me!

PR #6, opened 2020-07-23
Site layout/styling

kyracq:

Note: For resizing, I'm redrawing the map on every window resize since the GeoChart API doesn't allow dynamic sizes. This solution doesn't seem optimal since the map is constantly refreshing when you change the window size, but after trying some different options only this solution worked.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -33,8 +33,8 @@
         </form>
       </div>
       <div id="result">
-        <div id="map"></div>
         <div id="more-info"></div>
+        <div id="map"></div>
       </div>
     </div>
     <footer>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -5,33 +5,41 @@
   <meta charset="UTF-8">
   <title>Census Viz</title>
   <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
+  <link rel="stylesheet" href="style.css">
   <script src="script.js"></script>
 </head>
 
 <body onload="registerServiceWorker()">
   <main>
-    <div id="query">
-      <form id="query-form" onsubmit="passQuery(); return false;">
-        <p>
-          How many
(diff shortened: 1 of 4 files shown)


src/main/webapp/style.css
```diff
@@ -13,17 +13,54 @@ select:focus {
...
+  line-height: 1em;
+  padding-left: 30px;
+}
+
+#query {
+  text-align: center;
+}
+
+input[type=submit] {
+  background: #bfbfbf;
```
nsri-google, line 31: This makes the button seem grayed out which makes seem like it is not clickable.

kyracq: Changed button color to green.

src/main/webapp/index.html
```diff
@@ -5,33 +5,41 @@
...
+              <option value="county">county</option>
+              <option value="state">state</option>
+            </select>
+            ?
+          <input type="submit" value="go">
+        </form>
+      </div>
+      <div id="result">
+        <div id="map"></div>
+        <div id="more-info"></div>
```
shulijones, line 37: I'd like to put more-info above the map, I think. Since the map is so big, you have to scroll down to see more-info and don't notice it's there.

src/main/webapp/script.js
```diff
@@ -24,47 +24,51 @@ function passQuery() {
...
+    region: 'US',
+    resolution: 'provinces',
   };
-  var chart = new google.visualization.GeoChart(document.getElementById('map'));
+  document.getElementById('result').style.display = 'block';
+  const chart = new google.visualization.GeoChart(document.getElementById('map'));
   chart.draw(data, options);
-  document.getElementById('more-info').innerHTML = 'These populations are divided by 1000';
+  document.getElementById('more-info').innerHTML =
+      'These populations are divided by 1000';
```
nsri-google, line 68: As Anirudh suggested elsewhere this should be something like "values in 1000s"

PR #11, opened 2020-07-24
Testing fix for JS

kyracq:
Turned off indentation checking for JavaScript. To auto fix linting issues for JS, you'll need npm to install eslint. You can install npm with "sudo apt-get install npm", then install eslint with "sudo npm install -g eslint". 

The google java formatter doesn't allow configuring the linter, so right now I have it printing the correctly formatted file in the test output. I fixed Java formatting issues by copying the corrected file into VSCode. The selecting in the test output is finicky, if you scroll it unselects what you'd selected. I can also completely remove the Java linter if that'd be simpler.

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -19,69 +19,75 @@ google.charts.load('current', {
 });
 
 function passQuery() {
-  document.getElementById('result').style.display = 'block';
+  const information = document.getElementById('more-info');
+  information.innerText = 'Please wait. Loading...';
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
+  const action = query.get('action');
   const location = query.get('location');
 
-  fetch('/query?person-type=' + personType + '&location=' + location)
-      .then((response) => {
-        if (response.ok) {
-          response.json().then((jsonResponse) => JSON.parse(jsonResponse))
-              .then((data) => {
-                // data is a 2D array, where the first row is a header row and all
-                // subsequent rows are one piece of data (e.g. for a state or county)
-                displayVisualization(data);
-              });
-        } else {
-          console.log('There was an error');
-        }
-      });
+  const fetchUrl = '/query?person-type=' + personType +
+      '&action=' + action +
+      '&location=' + location;
+  
+  fetch(fetchUrl)
+    .then((response) => {
+      if (response.ok) {
+        response.json().then((jsonResponse) => JSON.parse(jsonResponse))
+        .then((data) => {
+          // data is a 2D array, where the first row is a header row and all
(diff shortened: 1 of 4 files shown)


src/main/webapp/script.js
```diff
@@ -19,75 +19,69 @@ google.charts.load('current', {
 });
 
 function passQuery() {
-  const information = document.getElementById('more-info');
-  information.innerText = 'Please wait. Loading...';
```
shulijones, line 23: Is there a reason/did you mean to delete the part that adds the loading text? Not sure if you implemented a similar thing elsewhere, but my original thought was that showing the result div (or even just the more-info div) now would allow that loading text to be seen by the user. If there's no loading text, no need to make the div visible.

src/main/webapp/script.js
```diff
@@ -19,75 +19,69 @@ google.charts.load('current', {
...
-          // subsequent rows are one piece of data (e.g. for a state or county)
-          information.innerText = '';
-          displayVisualization(data);
-        });
-      } else {
-        console.log(
-          `An error occurred: ${response.status}: ${response.statusText}`);
-      }
-    });
+  fetch('/query?person-type=' + personType + '&location=' + location)
```
shulijones, line 27: Oh wait. It looks like you weren't up to date with master -- this is a revert to before I added more dropdown options.

src/main/webapp/script.js
kyracq: Yeah, no I didn't mean to delete that, when I look at master I don't see these lines though, you might have to merge your branch into master first. I started this branch directly from master

src/main/webapp/script.js
```diff
@@ -19,75 +19,69 @@ google.charts.load('current', {
...
+// census API. The function creates the necessary visualization with the census data.
 function displayVisualization(censusDataArray) {
   if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
     google.charts.setOnLoadCallback(drawRegionsMap(censusDataArray));
   } else {
     // We currently do not have counties implemented
-    const errorMessage = 'We do not support this visualization yet';
     document.getElementById('map').innerHTML = '';
-    document.getElementById('more-info').innerHTML = errorMessage;
+    document.getElementById('more-info').innerHTML = 'We do not have this information yet.';
```
otobin, line 50: It looks like some of the deletions/insertions are up to date but some aren't. Here it should be 'We do not support this visualization yet' but later in the code it shows more up to date insertions.

kyracq: I must have mistakenly started from an old version of master I had locally, fixed to match master

PR #13, opened 2020-07-27
Add title to map

kyracq:

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -23,17 +23,17 @@ function passQuery() {
   document.getElementById('result').style.display = 'block';
   const information = document.getElementById('more-info');
   information.innerText = 'Please wait. Loading...';
-  
+
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
   const location = query.get('location');
-  
+
   const region =
       location === 'state' ? 'U.S. state' : 'State Name county';
   const title = 'Population who ' + action +
       ' in each ' + region + ' (' + personType.replace('-', ' ') + ')';
-  
+
   const fetchUrl = '/query?person-type=' + personType +
       '&action=' + action +
       '&location=' + location;
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -19,6 +19,7 @@ google.charts.load('current', {
 });
 
 function passQuery() {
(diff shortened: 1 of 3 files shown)


src/main/webapp/script.js
```diff
@@ -19,13 +19,18 @@ google.charts.load('current', {
...
   const information = document.getElementById('more-info');
   information.innerText = 'Please wait. Loading...';
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
   const location = query.get('location');
+  const region =
+      location === 'state' ? 'U.S. state' : 'State Name county';
+  const title = 'Population who ' + action +
+      ' in each ' + region + ' (' + personType + ')';
```
shulijones, line 33: For increased readability, maybe do a find and replace for converting dashes to spaces when making the title? (so that eg all-ages appears as all ages)

src/main/webapp/script.js
```diff
@@ -73,6 +78,7 @@ function drawRegionsMap(censusDataArray) {
     'resolution': 'provinces',
     'colorAxis': {colors: ['white', 'blue']},
   };
+  document.getElementById('map-title').innerHTML = title;
```
shulijones, line 81: Is there a reason to use innerHTML instead of innertText, since the title is just regular text?

src/main/webapp/script.js
```diff
@@ -19,13 +19,18 @@ google.charts.load('current', {
...
 function passQuery() {
+  document.getElementById('map-title').innerHTML = '';
   document.getElementById('result').style.display = 'block';
   const information = document.getElementById('more-info');
   information.innerText = 'Please wait. Loading...';
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
   const location = query.get('location');
+  const region =
```
shulijones, line 30: Nit but I might add an empty line here, and maybe one more somewhere above, for increased readability

src/main/webapp/script.js
kyracq: Done!

src/main/webapp/script.js
kyracq: Switched to innerText

src/main/webapp/script.js
kyracq: Done!

PR #14, opened 2020-07-27
Percentage function

otobin:
When a request has both a percentage and total, the percent of the total is shown.

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -53,7 +53,9 @@ function passQuery() {
 // Eg: Header for the query that finds for populations of all counties is
 // ["NAME","S0201_001E","state","county"]
 function displayVisualization(censusDataArray) {
-  if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
+  console.log(censusDataArray[0]);
+  const dataLength = censusDataArray[0].length;
+  if (censusDataArray[0][length - 1] != 'county') {
     google.charts.setOnLoadCallback(drawRegionsMap(censusDataArray));
   } else {
     // We currently do not have counties implemented
@@ -85,20 +87,21 @@ function createDataArray(censusDataArray) {
   const vizDataArray = []
   if (censusDataArray[0].length == 4) {
     censusDataArray.forEach((state) => {
-      vizDataArray.push([state[0], state[1]/1000, state[2]]);
+      vizDataArray.push([state[0], percentToTotal(state[1], state[2])]);
     });
-    // Changes the header of the vizDataArray to match the Visualization API
-    vizDataArray[0][0] = 'State';
-    vizDataArray[0][1] = 'Population';
-    vizDataArray[0][2] = 'Percentage';
-    return vizDataArray;
   } else {
     censusDataArray.forEach((state) => {
       vizDataArray.push([state[0], state[1]/1000]);
     });
-    // Changes the header of the vizDataArray to match the Visualization API
-    vizDataArray[0][0] = 'State';
-    vizDataArray[0][1] = 'Population';
-    return vizDataArray;
   }
+  // Changes the header of the vizDataArray to match the Visualization API
+  vizDataArray[0][0] = 'State';
+  vizDataArray[0][1] = 'Population';
+  return vizDataArray;
+}
+
+// percentToTotal takes in the total number of people in a category and the percentage and 
+// returns the total 
+function percentToTotal(totalNumber, percentage) {
+  return (totalNumber/100) * percentage;
 }
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -28,7 +28,12 @@ function passQuery() {
   const personType = query.get('person-type');
   const action = query.get('action');
   const location = query.get('location');
-  const description = {age: personType, action: action};
+  const actionToText = new Map();
+  actionToText['moved'] = 'moved to';
+  if (actionToText.has(action)) {
+    action = actionToText[action];
+  } 
+  const description = {action: action, age: personType};
 
   const region =
       location === 'state' ? 'U.S. state' : 'State Name county';
@@ -96,11 +101,8 @@ function drawRegionsMap(censusDataArray, description, title) {
 // and reformats it into a data table for the visualization API.
 function createDataArray(censusDataArray, description) {
   const vizDataArray = [];
-  // If the length of the census data array is four, the census call has
-  // generated a percentage and total population so we do a calculation
-  //  to find the % of the total.If the length is not four then we can
-  // just return given population.
-  if (censusDataArray[0].length === 4) {
+  // Check to see if an extra calculation for percentages is needed
+  if (checkPercentage(censusDataArray[0])) {
     censusDataArray.forEach((state) => {
       vizDataArray.push([state[0], percentToTotal(state[1], state[2])]);
     });
@@ -125,6 +127,25 @@ function createDataArray(censusDataArray, description) {
 // and the percentage and returns the total
 function percentToTotal(totalNumber, percentage) {
   return (totalNumber/100) * percentage;
+}
+
+// checkPercentage() Takes in the header of a census query and returns 
+// whether or not the total needs to be calculated using the percentToTotal() function
+function checkPercentage(headerColumn) {
+  // percentageQueries is a list of queries that return percents and not raw
+  // data. It is hard coded for now.
+  const percentageQueries = ['S0201_157E', 'S0201_126E'];
+  let i;
+  // Iterate through all the data variables in the header.
+  // Eg: ["NAME","S0201_119E","S0201_126E","state"]
+  // We need to return true
+  for (i = 1; i < headerColumn.length - 1; i++) {
+    // if the current data is in the percentage array
+    if (percentageQueries.indexOf(headerColumn[i]) > -1) {
+      return true;
+    }
+  }
+  return false;
 }
 
 function resizeVisualization() {
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -53,9 +53,8 @@ function passQuery() {
 // Eg: Header for the query that finds for populations of all counties is
 // ["NAME","S0201_001E","state","county"]
 function displayVisualization(censusDataArray) {
-  console.log(censusDataArray[0]);
   const dataLength = censusDataArray[0].length;
-  if (censusDataArray[0][length - 1] != 'county') {
+  if (censusDataArray[0][dataLength - 1] != 'county') {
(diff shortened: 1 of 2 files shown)


src/main/webapp/script.js
```diff
@@ -53,7 +53,9 @@ function passQuery() {
 // Eg: Header for the query that finds for populations of all counties is
 // ["NAME","S0201_001E","state","county"]
 function displayVisualization(censusDataArray) {
-  if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
+  console.log(censusDataArray[0]);
```
shulijones, line 56: If this was for debugging, probably want to delete it now

src/main/webapp/script.js
```diff
@@ -53,7 +53,9 @@ function passQuery() {
...
+  const dataLength = censusDataArray[0].length;
```
shulijones, line 57: You don't seem to use this variable anywhere that I can see

src/main/webapp/script.js
```diff
@@ -82,12 +84,24 @@ function drawRegionsMap(censusDataArray) {
 // createDataArray takes in the data array returned by the census API
 // and reformats it into a data table for the visualization API.
 function createDataArray(censusDataArray) {
-  const vizDataArray = [];
-  censusDataArray.forEach((state) => {
-    vizDataArray.push([state[0], state[1]/1000]);
-  });
+  const vizDataArray = []
+  if (censusDataArray[0].length == 4) {
```
shulijones, line 88: I might add a quick comment here explaining a bit more about the two types of data you're differentiating between with the if/else

src/main/webapp/script.js
```diff
@@ -82,12 +84,24 @@ function drawRegionsMap(censusDataArray) {
...
+      vizDataArray.push([state[0], percentToTotal(state[1], state[2])]);
+    });
+  } else {
+    censusDataArray.forEach((state) => {
+      vizDataArray.push([state[0], state[1]/1000]);
+    });
+  }
   // Changes the header of the vizDataArray to match the Visualization API
   vizDataArray[0][0] = 'State';
   vizDataArray[0][1] = 'Population';
```
shulijones, line 99: It's not always population anymore so we will need to change this (maybe just hardcode by using a small dictionary, and having its keys be the possible `action` strings from the form?)

src/main/webapp/script.js
```diff
@@ -53,7 +53,9 @@ function passQuery() {
 // Eg: Header for the query that finds for populations of all counties is
 // ["NAME","S0201_001E","state","county"]
 function displayVisualization(censusDataArray) {
-  if (censusDataArray[0][2] == 'state' && censusDataArray[0][3] != 'county') {
+  console.log(censusDataArray[0]);
+  const dataLength = censusDataArray[0].length;
+  if (censusDataArray[0][length - 1] != 'county') {
```
kyracq, line 58: Style guide recommends using === and !== instead of == and != (https://google.github.io/styleguide/jsguide.html#features-equality-checks)

src/main/webapp/script.js
kyracq, line 88: Like I mentioned in another comment should be triple equals

src/main/webapp/script.js
```diff
@@ -82,12 +84,24 @@ function drawRegionsMap(censusDataArray) {
...
 // and reformats it into a data table for the visualization API.
 function createDataArray(censusDataArray) {
-  const vizDataArray = [];
-  censusDataArray.forEach((state) => {
-    vizDataArray.push([state[0], state[1]/1000]);
-  });
+  const vizDataArray = []
+  if (censusDataArray[0].length == 4) {
+    censusDataArray.forEach((state) => {
+      vizDataArray.push([state[0], percentToTotal(state[1], state[2])]);
```
kyracq, line 90: Just a suggestion: might be good to add both percent and number and allow user to choose which to display?

otobin: Unfortunately the data visualization api doesn't allow for both to be shown at the same time so we'll have to add another dropdown menu later

src/main/webapp/script.js
otobin: fixed!

src/main/webapp/script.js
otobin: true! Just added it

src/main/webapp/script.js
otobin: Done! Not very grammatically correct but should do the trick for now

src/main/webapp/script.js
otobin: fixed!

src/main/webapp/script.js
otobin: thank you for the reminder! I always forget this

src/main/webapp/script.js
```diff
@@ -90,17 +94,39 @@ function drawRegionsMap(censusDataArray, title) {
...
   const vizDataArray = [];
-  censusDataArray.forEach((state) => {
-    vizDataArray.push([state[0], state[1]/1000]);
-  });
-  // Changes the header of the vizDataArray to match the Visualization API
+  // If the length of the census data array is four, the census call has
+  // generated a percentage and total population so we do a calculation
+  //  to find the % of the total.If the length is not four then we can
+  // just return given population.
+  if (censusDataArray[0].length === 4) {
```
nsri-google, line 103: this is very brittle. Is there some other way (like test in the column header) to identify what different columns are?

src/main/webapp/script.js
```diff
@@ -90,17 +94,39 @@ function drawRegionsMap(censusDataArray, title) {
...
+    });
+  }
+  // Changes the header of the vizDataArray to match Visualization API
   vizDataArray[0][0] = 'State';
-  vizDataArray[0][1] = 'Population';
+  // Add a more accurate descriptor for each state
+  vizDataArray[0][1] = 'Population of people ' +
+    description.age + ' who ' + description.action;
+  // make it moved to
+  if (description.action === 'moved') {
```
nsri-google, line 118: again this feels hacky. Maybe add a map from action to text to display. if an entry is found for a specific action use the text from the map or use the action as is.

src/main/webapp/script.js
```diff
@@ -90,17 +94,39 @@ function drawRegionsMap(censusDataArray, title) {
...
 // createDataArray takes in the data array returned by the census API
 // and reformats it into a data table for the visualization API.
-function createDataArray(censusDataArray) {
+function createDataArray(censusDataArray, description) {
   const vizDataArray = [];
-  censusDataArray.forEach((state) => {
-    vizDataArray.push([state[0], state[1]/1000]);
-  });
-  // Changes the header of the vizDataArray to match the Visualization API
+  // If the length of the census data array is four, the census call has
```
nsri-google, line 99: Is there a website link where you found this? Can you add that link here?

otobin: Right now our website is programmed to take in states and counties. A county call has the header ["NAME","S0201_001E","state","county"] while a state call can either have the header ["NAME","S0201_154E","S0201_157E","state"] or ["NAME","S0201_154E","state"] depending on the type of call that is made. This isn't from a website it's just from my experience working with the limited queries thus far. It's also not a sure fire way to test for a percentage query, but it works with our limited set of queries.

src/main/webapp/script.js
otobin: Could you elaborate on what you mean by test in the column header?

 In the displayVisualization function, we test to make sure that the query is not a county function so we can return the appropriate error response. County queries and percentage queries are the only types of queries we're dealing with that have four columns, so if the query has four columns here then it must be a percentage query. There are other ways to test for it at this point, like creating a list of known census data id's and putting them into a map or array. Is that what you mean? This would also be repeated logic from the javascript backend and would also be just as hacky given we have no formal algorithm right now.

nsri-google: sorry wrong choice of words. I meant if we could determine from the header which column is which

otobin: Yes, we could keep track of what headers correspond to percentage calls. Eg: 157E and 126E are headers that correspond percentage calls for working and moving. 

We could keep these in an array and then check to see if any of the headers appear in the array. It would be more time consuming and still hacky because we're hard coding the values but more maintainable to build on.

src/main/webapp/script.js
nsri-google: I am little confused. I thought you specify the variables you want in the request and the api returns appropriate values. Can you give me example queries where it returns 4 columns vs when it returns 3 columns.

src/main/webapp/script.js
nsri-google: If you move that into separate mapping, it can be hard coded now but later maybe we can find a way to get that information directly from the census data api. Also it is not depending on number of columns or order of columns which makes it a little more robust.

src/main/webapp/script.js
otobin: A query for "People of all ages who moved to each state" has the header ["NAME","S0201_119E","S0201_126E","state"]. We must do a calculation using the second and third columns of data to return the appropriate total. However, a query for "People of all ages who live in each state" has the header ["Name", "S0201_001E", "state"]. 

This is why I was testing the length, because the number of columns will indicate whether we have to do a percent to total calculation.

nsri-google: ok it seems to be this should actually be happening in the backend which has all the information about the tables and fields etc. and the frontend should not have to do these kind manipulations. Let's talk about this in one of our meetings.

src/main/webapp/script.js
otobin: Fixed!

src/main/webapp/script.js
otobin: Sounds good!

src/main/webapp/script.js
otobin: Fixed to be hopefully not as hacky

src/main/webapp/script.js
```diff
@@ -90,17 +99,55 @@ function drawRegionsMap(censusDataArray, title) {
...
+  vizDataArray[0][1] = 'Population of people ' +
+    description.age + ' who ' + description.action;
+  // make it moved to
+  if (description.action === 'moved') {
+    vizDataArray[0][1] += ' to';
+  }
   return vizDataArray;
 }
 
+// percentToTotal takes in the total number of people in a category
```
nsri-google, line 126: This is ok for now but I think the cleaner solution is to have the backend put the headers of the form ('population', 'percentage' etc.) and the front does not need to know about things like 'S0201_157E', 'S0201_126E'

otobin: Got it. The three of us can discuss this soon.

PR #15, opened 2020-07-27
Switch location dropdown to write-in (datalist)

shulijones:
The location dropdown is now an HTML datalist instead of a select. That means that the user can write in the box, and the dropdown will show them only the options containing the string they've written (e.g. "n" shows New Jersey and California, but "ne" only shows New Jersey). 

I also updated the dropdown so that instead of having "every county" and "every state", it has "every state", "California", and "New Jersey". For now, choosing CA or NJ will get you the same not-yet message as choosing every county, but my thought was that we could use one or both of those states as a test run for amcharts.

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -22,10 +22,13 @@ function passQuery() {
   document.getElementById('result').style.display = 'block';
   const information = document.getElementById('more-info');
   information.innerText = 'Please wait. Loading...';
+
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
-  const location = query.get('location');
+  const locationName = query.get('location');
+  const location = document.querySelector("#location option[value='"+locationName+"']").dataset.value;
+
   const fetchUrl = '/query?person-type=' + personType +
       '&action=' + action +
       '&location=' + location;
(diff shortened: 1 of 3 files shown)


src/main/webapp/script.js
```diff
@@ -22,10 +22,13 @@ function passQuery() {
   document.getElementById('result').style.display = 'block';
   const information = document.getElementById('more-info');
   information.innerText = 'Please wait. Loading...';
+
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
-  const location = query.get('location');
+  const locationName = query.get('location');
+  const location = document.querySelector("#location option[value='"+locationName+"']").dataset.value;
```
nsri-google, line 30: I think with datalist you can type a value not in the list so you need some error checking here

nsri-google: see https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_datalist, you can type random text in the box and submit and it will still send it through

shulijones: This is true, but I disabled that functionality in the HTML. The input of type list (that the datalist belongs to) has a pattern defined, so it only accepts answers that fit the pattern. The pattern is just all of the possible answers to the datalist separated by | (or symbols). It's a bit inelegant, so definitely open to alternate suggestions, but I do like that it rejects incorrect answers in the front end itself.

PR #16, opened 2020-07-28
Display all errors with a relevant message

shulijones:
Display all errors on the front end with a status code and message about what the error is.

I added a (for now, very simple) JS function to start to make error handling more robust for the future as well. If the census server breaks in some way while the data is being fetched, the user will see "Error 403: An error occurred while trying to retrieve census data." If they submit a query for any visualization we don't have right now, they'll see "Error 400: We do not support this visualization yet."

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -28,7 +28,8 @@ function passQuery() {
   const action = query.get('action');
 
   const locationName = query.get('location');
-  const location = document.querySelector("#location option[value='"+locationName+"']").dataset.value;
+  const location = document.querySelector(
+    "#location option[value='"+locationName+"']").dataset.value;
 
   const region =
       location === 'state' ? 'U.S. state' : 'State Name county';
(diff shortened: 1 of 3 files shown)


src/main/webapp/script.js
```diff
@@ -20,16 +20,16 @@ google.charts.load('current', {
...
-  information.innerText = 'Please wait. Loading...';
 
   const query = new FormData(document.getElementById('query-form'));
   const personType = query.get('person-type');
   const action = query.get('action');
 
   const locationName = query.get('location');
-  const location = document.querySelector("#location option[value='"+locationName+"']").dataset.value;
+  const location = document.querySelector(
+    "#location option[value='"+locationName+"']").dataset.value;
```
kyracq, line 32: Single quotes to match style guide

shulijones: I think it has to be double quotes because I want the string that's created to have single quotes inside of it (the value='locationName' part)

kyracq: I think you can use single quotes in a single quote string with a backslash: '#location option[value=\''+locationName+'\']'

shulijones: That worked, thanks!

PR #17, opened 2020-07-28
Amcharts

kyracq:
- Display county maps for NJ and CA using amcharts
- Switched US map to amcharts
- Disabled zoom in with mouse scroll, added buttons instead
- Zooms in when you click on a state/county (Home button to zoom all the way out)

~~Currently when you first open the site the response from the query servlet doesn't match the query until after empty cache. For example this county query: https://censusviz.appspot.com/query?person-type=under-18&action=live&location=county&state-number=34 shows by state data. Let me know if you also come across this issue or know of how to resolve it~~
To get around this issue I currently have the NJ and CA queries hardcoded in.

### DIFF
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -24,6 +24,9 @@ public class QueryServlet extends HttpServlet {
           ImmutableMap.of("over-18", "154E,S0201_157E"),
           "moved",
           ImmutableMap.of("all-ages", "119E,S0201_126E"));
+  
+  Map<String, String> locationTypeToQuery =
+      ImmutableMap.of("county", "county:*&in=state:", "state", "state:*");
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
@@ -39,19 +42,13 @@ public class QueryServlet extends HttpServlet {
       return;
     }
 
-    String locationQuery;
-    if (location == "county" && stateNumber != null) {
-      locationQuery = "county:*&in=state:" + stateNumber;
-    } else {
-      locationQuery = "state:*";
-    }
-    
     URL fetchUrl =
         new URL(
             "https://api.census.gov/data/2018/acs/acs1/spp?get=NAME,S0201_"
                 + queryToDataRow.get(action).get(personType)
                 + "&for="
-                + locationQuery
+                + locationTypeToQuery.get(location)
+                + stateNumber
                 + "&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
 
     HttpURLConnection connection = (HttpURLConnection) fetchUrl.openConnection();
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -30,6 +30,7 @@ public class QueryServlet extends HttpServlet {
     String personType = request.getParameter("person-type");
     String action = request.getParameter("action");
     String location = request.getParameter("location");
+    String stateNumber = request.getParameter("state-number");
 
     if (!queryToDataRow.containsKey(action)
         || !queryToDataRow.get(action).containsKey(personType)) {
@@ -38,13 +39,20 @@ public class QueryServlet extends HttpServlet {
       return;
     }
 
+    String locationQuery;
+    if (location == "county" && stateNumber != null) {
+      locationQuery = "county:*&in=state:" + stateNumber;
+    } else {
+      locationQuery = "state:*";
+    }
+    
     URL fetchUrl =
         new URL(
             "https://api.census.gov/data/2018/acs/acs1/spp?get=NAME,S0201_"
                 + queryToDataRow.get(action).get(personType)
                 + "&for="
-                + location
-                + ":*&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
+                + locationQuery
+                + "&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
 
     HttpURLConnection connection = (HttpURLConnection) fetchUrl.openConnection();
     connection.setRequestMethod("GET");
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -26,14 +26,16 @@ public class QueryServlet extends HttpServlet {
           ImmutableMap.of("all-ages", "119E,S0201_126E"));
 
   Map<String, String> locationTypeToQuery =
-      ImmutableMap.of("county", "county:*&in=state:", "state", "state:*");
+      ImmutableMap.of(
+        "06", "county:*&in=state:06",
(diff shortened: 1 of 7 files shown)


src/main/webapp/index.html
```diff
@@ -30,9 +35,9 @@ <h1>Census Viz</h1>
           </select>
           <input type="list" list="location" name="location" pattern="every US state|California|New Jersey">
             <datalist id="location">
-              <option data-value="state" value="every US state">
-              <option data-value="county" value="California">
-              <option data-value="county" value="New Jersey">
+              <option data-value="state" value="every US state"></option>
```
shulijones, line 38: Nit: I think the convention for datalist is to not close option tags, eg they don't here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

src/main/java/QueryServlet.java
```diff
@@ -43,8 +47,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
             "https://api.census.gov/data/2018/acs/acs1/spp?get=NAME,S0201_"
                 + queryToDataRow.get(action).get(personType)
                 + "&for="
-                + location
-                + ":*&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
+                + locationTypeToQuery.get(location)
+                + stateNumber
```
shulijones, line 51: It looks like when the query is for all states, you set the state-number parameter to an empty string in the JS so that it can always be appended here. This definitely works, but I wonder if it would make more sense from an error handling/avoidance perspective to do it a bit differently? For example, not including the state-number parameter if we didn't ask for states, or making it equal to false or something, and then only appending the stateNumber string if there was an actual number.

src/main/webapp/script.js
```diff
@@ -56,14 +82,84 @@ function passQuery() {
...
+    logarithmic: true,
+  });
+  polygonSeries.useGeodata = true;
+
+  const countyData = data.map((value) => {
+      return {id: value[2] + value[3], value: value[1]};
+  });
+  polygonSeries.data = countyData;
+
+  // Set up heat legend
```
shulijones, line 110: This could be something for later, but the GeoCharts heat legend and the AmCharts heat legend look very similar but are positioned very differently on the page. It would be nice to have them appearing in the same spot.

src/main/webapp/script.js
```diff
@@ -73,7 +169,7 @@ function displayVisualization(censusDataArray, title) {
 }
 
 // Takes in a 2D array from the census API and displays the visualization
-function drawRegionsMap(censusDataArray, title) {
+function drawRegionsMap(censusDataArray) {
```
otobin, line 172: I think this function should be drawRegionsMap(censusDataArray, title) because you still use the title in the function

src/main/webapp/script.js
```diff
@@ -29,16 +41,25 @@ function passQuery() {
...
+  const location = document.querySelector(
+      '#location option[value=\''+locationName+'\']').dataset.value;
 
-  const region =
-      location === 'state' ? 'U.S. state' : 'State Name county';
+  const isCountyQuery = location !== 'state';
+  const region = isCountyQuery ? locationName + ' county' : 'U.S. state';
   const title = 'Population who ' + action +
       ' in each ' + region + ' (' + personType.replace('-', ' ') + ')';
+  document.getElementById('map-title').innerText = title;
```
otobin, line 51: I like that you moved the title to here and not the displayVisualization function!

src/main/webapp/script.js
```diff
@@ -56,14 +82,84 @@ function passQuery() {
     });
 }
 
+function displayAmChartsMap(data, stateName) {
```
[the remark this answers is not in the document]
otobin: I think it makes more sense to do the data checking in displayVisualization() so that this function isn't even called if the data returned is empty

src/main/webapp/index.html
```diff
@@ -30,9 +35,9 @@ <h1>Census Viz</h1>
...
+              <option data-value="county" value="California"></option>
+              <option data-value="county" value="New Jersey"></option>
               <!-- will change latter two to 06 and 31 (state numbers) when counties are ready -->
```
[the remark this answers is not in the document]
kyracq: Thanks for this suggestion, done!

src/main/webapp/index.html
kyracq: Thanks, will have to figure out how let the linter allow this

src/main/java/QueryServlet.java
kyracq: Yeah, I agree the way I had it before didn't make much sense. For now I have the values for CA and NJ hardcoded in, will have to figure out how to do this dynamically for each state

src/main/webapp/script.js
```diff
@@ -30,43 +27,134 @@ function passQuery() {
...
+    } else {
+      heatLegend.valueAxis.hideTooltip();
+    }
+  }
+  polygonSeries.mapPolygons.template.events.on('out', function(event) {
+    heatLegend.valueAxis.hideTooltip();
+  });
+}
+
 function validateInput(dataListId) {
```
shulijones, line 158: I think you might have accidentally deleted this function's description (on the old lines 68 and 69).

src/main/webapp/script.js
```diff
@@ -30,43 +27,134 @@ function passQuery() {
...
+  }
+  polygonTemplate.tooltipText = '{name}\n' + descriptionString + ': {value}';
+  polygonTemplate.nonScalingStroke = true;
+  polygonTemplate.strokeWidth = 0.5;
+
+  // Create hover state and set alternative fill color
+  const hs = polygonTemplate.states.create('hover');
+  hs.properties.fill = am4core.color('#3c5bdc');
+
+
```
shulijones, line 138: Nit: can delete one empty line

src/main/webapp/script.js
kyracq: Thanks for catching this!

src/main/webapp/script.js
kyracq: Done.

src/main/webapp/script.js
```diff
@@ -30,41 +27,133 @@ function passQuery() {
...
+  const chart = am4core.create('map', am4maps.MapChart);
+  chart.height = 550;
+  chart.zoomControl = new am4maps.ZoomControl();
+  // only allow zooming with buttons
+  chart.mouseWheelBehavior = 'none';
+
+  // Create map instance
+  if (isCountyQuery) {
+    chart.geodata = location ===
+        '06' ? am4geodata_region_usa_caLow : am4geodata_region_usa_njLow;
```
nsri-google, line 85: for now this is ok but as we add more states we should look to have a clean way to store this mapping. For example this could also be part of the stateinfo.

PR #19, opened 2020-07-29
Update location dropdown

shulijones:
New: all boxes are now datalists (with default inputs given).

Update location dropdown so that it alerts the user if they begin to type an option that does not exist, and so that it empties itself when the user returns to make a second query.

The location box will now turn red when you type a string that is not contained inside any of the datalist's options. Every time you click back into the datalist, it will empty itself and show you the full list of options again. 
I also added code to turn the autocomplete feature off that Sriram was seeing, but since I never saw it in the first place, I can't verify that it works. 
Lastly, it is no longer possible to submit with that field empty.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -19,8 +19,10 @@
         <p>
           How many
           <input type="list" list="person-type" name="person-type" id="person-type-list"
-              autocomplete="off" required onkeyup="validateInput('person-type', 'person-type-list')"
-              onfocus="this.value=''" value="people" pattern="children|adults|people">
+              autocomplete="off" required onkeyup="validateInput('person-type')"
+              onfocus="storeValueAndEmpty('person-type')" 
+              onfocusout="replaceValueIfEmpty('person-type')"
+              value="people" pattern="children|adults|people">
             <datalist id="person-type">
               <option data-value="under-18" value="children">
               <option data-value="over-18" value="adults">
@@ -29,8 +31,10 @@
           </input>
 
           <input type="list" list="action" name="action" id="action-list"
-              autocomplete="off" required onkeyup="validateInput('action', 'action-list')"
-              onfocus="this.value=''" value="live in" pattern="live in|work in|moved to" >
+              autocomplete="off" required onkeyup="validateInput('action')"
+              onfocus="storeValueAndEmpty('action')" 
+              onfocusout="replaceValueIfEmpty('action')"
+              value="live in" pattern="live in|work in|moved to" >
             <datalist id="action">
               <option data-value="live" value="live in">
               <option data-value="work" value="work in">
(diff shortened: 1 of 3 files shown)


src/main/webapp/index.html
```diff
@@ -28,7 +28,9 @@ <h1>Census Viz</h1>
             <option value="work">work in</option>
             <option value="moved">moved to</option>
           </select>
-          <input type="list" list="location" name="location" pattern="every US state|California|New Jersey">
+          <input type="list" list="location" name="location" id="locationlist"
+          autocomplete="off" required onkeyup="validateLocation()"
```
kyracq, line 32: I think these wrapped lines should be indented: https://google.github.io/styleguide/htmlcssguide.html#HTML_Line-Wrapping

src/main/webapp/script.js
```diff
@@ -60,6 +60,27 @@ function passQuery() {
     });
 }
 
+function validateLocation() {
+  const datalist = document.getElementById('location');
```
nsri-google, line 64: This function does not have to know about the 'location' and 'locationList'. Instead you can pass them as parameters from the html and here you method would be function validationLocation(locationVar, locationListVar)

src/main/webapp/script.js
```diff
@@ -60,6 +60,27 @@ function passQuery() {
...
+  const inputlist = document.getElementById('locationlist');
+
+  const options = datalist.options;
+  const typedSoFar = inputlist.value.toLowerCase();
+  let matchesAnOption = false;
+
+  for (const option of options) {
+    if (option.value.toLowerCase().includes(typedSoFar)) {
+      matchesAnOption = true;
+      inputlist.style.borderColor = '#767676'; // dark gray
```
nsri-google, line 74: a cleaner option is to defne css classes for invalid-input and valid-input and here you would just change the class to one or the other. This way you later change how you style invalid and valid input just by changing css and not touching any code.

src/main/webapp/script.js
```diff
@@ -60,6 +60,27 @@ function passQuery() {
...
+      inputlist.style.borderStyle = 'none none solid none';
+    }
```
nsri-google, line 76: At this point you already have match so you can return from this function. there is no need to compare with other options. And if you do that then you don't even need matchesAnOption variable. When you fall out of the for loop you know have not matched any options.

src/main/webapp/index.html
shulijones: Done, thanks!

src/main/webapp/script.js
shulijones: That is an excellent idea. Done

src/main/webapp/script.js
shulijones: Very good point. Done

src/main/webapp/script.js
shulijones: Done

PR #20, opened 2020-08-03
Am charts and maps

otobin:
Inserted geoJson and buttons to toggle between the two. 


src/main/webapp/script.js
```diff
@@ -14,6 +14,21 @@ function registerServiceWorker() {
   }
 }
 
+// Maps the numerical value of the state to data used to create
+// needed to create geoJson map of the state.
+const states = {
```
nsri-google, line 19: ok for now but we should move this into stateinfo js file

src/main/webapp/script.js
```diff
@@ -71,22 +87,66 @@ function passQuery() {
...
+  if (isCountyQuery) {
+    setStyleForCountyQuery();
+    const mapsData = getMapsData(censusDataArray);
+    const amChartsData = createDataArray(censusDataArray, isCountyQuery);
+    displayAmChartsMap(amChartsData, description, geoData);
+    displayCountyGeoJson(mapsData, location);
+  } else {
+    setStyleForStateQuery();
+    const amChartsData = createDataArray(censusDataArray, isCountyQuery);
+    displayAmChartsMap(amChartsData, description, geoData);
```
nsri-google, line 112: Do we not want to display geocharts for this option?

src/main/webapp/script.js
```diff
@@ -260,3 +295,137 @@ function checkPercentage(headerColumn) {
...
 }
+
+// Returns an object containing all of the relevant data in order
+// to render the geoJson map of the counties.
+function getMapsData(censusDataArray) {
+  const countyToPopMap = new Map();
+  const populationsList = [];
+  Array.prototype.forEach.call(censusDataArray, (county) => {
+    if (county[0] !== 'NAME') {
+      // The current county strings are in a layout like this:
```
nsri-google, line 306: this is very brittle. Is this the only way the census data identifies the county?

nsri-google: How does the amcharts display code handle this situation?

src/main/webapp/script.js
otobin: Great point! Done.

src/main/webapp/script.js
otobin: The three of us never discussed using geoJson for states because Google Maps has states by default. I will ask tomorrow if we want to also add a toggle function for a state query.

src/main/webapp/script.js
otobin: Fixed to .slice(1) in order to get rid of the header when traversing the census data.

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
+function getMapsData(censusDataArray) {
+  const countyToPopMap = new Map();
+  const populationsList = [];
+  // Get rid of the header
+  const censusArray = censusDataArray.slice(1);
+  Array.prototype.forEach.call(censusArray, (county) => {
+    // The current county strings are in a layout like this:
+    // "Contra Costa County, California"
+    // and we need to get them like this "Contra Costa"
+    const countyAndStateArray = county[0].split(',');
```
nsri-google, line 295: depending on a specific format of text is risky. We can submit this for now but discuss how we can me this more robust.

src/main/webapp/script.js
```diff
@@ -59,10 +59,11 @@ function passQuery() {
       if (response.ok) {
         response.json().then((jsonResponse) => JSON.parse(jsonResponse))
         .then((censusDataArray) => {
-          // data is a 2D array, where the first row is a header row and all
-          // subsequent rows are one piece of data (e.g. for a state or county)
-          const data = createDataArray(censusDataArray, isCountyQuery);
-          displayAmChartsMap(data, description, location, isCountyQuery);
+          // censusDatadata is a 2D array, where the first row is a
```
shulijones, line 62: censusDataArray

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
+    map.data.revertStyle();
+    let i;
+    for (i = 0; i < openInfoWindows.length; i++) {
+      openInfoWindows[i].close();
+    }
+  });
+}
+
+// Functions to toggle between amcharts and maps.
+function showGeoJson() {
```
shulijones, line 383: These two functions are very similar, maybe they could just be one function and take in some parameters to determine which is made visible.

src/main/webapp/script.js
```diff
@@ -71,22 +72,66 @@ function passQuery() {
...
+    displayCountyGeoJson(mapsData, location);
+  } else {
+    setStyleForStateQuery();
+    const amChartsData = createDataArray(censusDataArray, isCountyQuery);
+    displayAmChartsMap(amChartsData, description, geoData);
+  }
+  document.getElementById('more-info').innerText = '';
+}
 
+// Check that the input being written to a datalist can match one of its options
```
shulijones, line 102: This is a more general comment, but we now have quite a lot of different functions in this JS file. I think it might make sense to create some "sections" within the file (just by using comments) for various types of functions, so that input validation isn't all mixed up with data visualization, and so on.

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
+function displayCountyGeoJson(mapsData, stateName) {
+  const map = new google.maps.Map(document.getElementById('map'), {
+    zoom: stateInfo[stateName].zoomLevel,
+    center: {lat: stateInfo[stateName].lat, lng: stateInfo[stateName].lng},
+  });
+
+  const countyToPopMap = mapsData.map;
+  const maxPopulation = mapsData.maxValue;
+  const minPopulation = mapsData.minValue;
+  const f = chroma.scale(['white', 'blue']).domain([minPopulation,
```
shulijones, line 345: This could have a more descriptive variable name

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
+    });
+  });
+
+  const openInfoWindows = [];
+  map.data.addListener('mouseover', function(event) {
+    map.data.overrideStyle(event.feature, {
+      fillColor: '#00ffff',
+    });
+    const contentString = '<p>' + event.feature.j.name +
+    '<p>Population: ' + countyToPopMap[event.feature.j.name];
```
shulijones, line 364: We will want to make the "population" descriptor more flexible here depending on what is being displayed, but that can be something for later

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
 }
+
+// Returns an object containing all of the relevant data in order
+// to render the geoJson map of the counties.
+function getMapsData(censusDataArray) {
+  const countyToPopMap = new Map();
+  const populationsList = [];
+  // Get rid of the header
+  const censusArray = censusDataArray.slice(1);
+  Array.prototype.forEach.call(censusArray, (county) => {
```
kyracq, line 291: I think this can be shortened to censusArray.forEach((county) => { ... }) like done here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/main/webapp/script.js
```diff
@@ -260,3 +280,137 @@ function checkPercentage(headerColumn) {
...
+function showAmCharts() {
+  const mapElement = document.getElementById('map');
+  const amChartsElement = document.getElementById('am-charts');
+  mapElement.style.display = 'none';
+  amChartsElement.style.display = 'block';
+}
+
+// Sets up the webpage for a county query. Displays buttons and
+// sets amcharts as the default visible map.
+function setStyleForCountyQuery() {
```
kyracq, line 399: Similar to Shuli's suggestion above, this and setStyleForStateQuery could be one function, where you check if the buttons are displayed remove them if they are etc.

src/main/webapp/script.js
anicr7: I think you can pass in a parameter and then decide which to display and which to remove.

src/main/webapp/script.js
```diff
@@ -71,22 +72,66 @@ function passQuery() {
...
-    chart.geodata = am4geodata_usaLow;
+    return am4geodata_usaLow;
   }
+}
+
+
+function displayVisualization(censusDataArray, description,
+  location, isCountyQuery) {
+  const geoData = getGeoData(location, isCountyQuery);
+  if (isCountyQuery) {
```
anicr7, line 88: This can be much simpler if the underlying functions like SetStyle can handle both cases.

src/main/webapp/script.js
otobin: Kyra pointed out that there is data in the amCharts geoJson that I can use.

src/main/webapp/script.js
otobin: nice catch

src/main/webapp/script.js
otobin: Fixed to a toggle() function that takes in parameters.

src/main/webapp/script.js
otobin: Very true! Will merge and then we can work on breaking it up.

src/main/webapp/script.js
otobin: changed name to colorScale

src/main/webapp/script.js
otobin: Fixed!

src/main/webapp/script.js
otobin: For some reason censusArray.forEach() was throwing an error earlier but I changed it back and now it works.

src/main/webapp/script.js
otobin: fixed to one function setStyle(isCountyQuery);

src/main/webapp/script.js
otobin: fixed to one function setStyle(isCountyQuery)

PR #22, opened 2020-08-04
Split script.js into two JS files

shulijones:
There is no change to the website's functioning, but script.js is now two files, one for handling the user's query and one for creating and displaying the visualizations.


src/main/webapp/index.html
```diff
@@ -14,10 +14,12 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/chroma-js/1.3.4/chroma.min.js"></script>
   <script src="state-info.js"></script>
   <link rel="stylesheet" href="style.css">
-  <script src="script.js"></script>
```
nsri-google, line 17: Is there a reason your are moving this from head to body. I think putting it in head is general practice. The other option is to put thinks at the end of body so that all html has rendered but in your case I don't think you can do that since the html refers to functions in the javascript.

If you found any references that suggest doing this I would love to find out more.

shulijones: The two files reference each other a bit, because the query JS file calls the dataviz JS file. I was looking up how to make sure that the function calls will work and I found this on stack overflow, which says that you can only have one JS file call another's functions by name if you put them in the body and not the head. https://stackoverflow.com/questions/25962958/calling-a-javascript-function-in-another-js-file

src/main/webapp/index.html
```diff
@@ -14,10 +14,12 @@
...
+  
 </head>
 
 <body onload="registerServiceWorker()">
+  <script src="viz-functions.js"></script>
```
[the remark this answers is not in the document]
shulijones: Good point, will change.

src/main/webapp/index.html
nsri-google: Are you actually running into this issue? when you add them to head it just loads all the functions (symbols) into memory. Only when it is actually called/executed (when you click on things in the HTML page) they will actually be evaluated but at that point all functions should be memory.

shulijones: You're right, it still works for me when I move it back to head. I can push that.

PR #23, opened 2020-08-04
All states

kyracq:

- All states work for AmCharts
- States dropdown created in for loop
- geoJson files only loaded when state queried


src/main/webapp/data-visualization.js
```diff
@@ -1,15 +1,27 @@
-function getGeoData(location, isCountyQuery) {
+async function getGeoData(location, isCountyQuery) {
```
shulijones, line 1: Could you add docstrings/descriptions to functions that don't have them?

src/main/webapp/state-info.js
```diff
@@ -18,9 +18,9 @@ const stateInfo = {
     '06': {
         'name': 'California',
         'ISO': 'US-CA',
-        lat: 38,
-        lng: -120,
-        zoomLevel: 6,
+        'lat': 38,
```
otobin, line 21: Thanks for fixing these! I'll make sure to add them for all the states so that the geoJson can load for all states.

src/main/webapp/query-control.js
```diff
@@ -119,4 +119,48 @@ function replaceValueIfEmpty(dataListId) {
...
+}
+
+function createStateDropdownList() {
+  const datalist = document.getElementById('location');
+  let optionElem = document.createElement('option');
+  optionElem.value = 'each U.S. state';
+  optionElem.setAttribute('data-value', 'state');
+  datalist.appendChild(optionElem);
+
+  for (const state in stateInfo) {
```
shulijones, line 155: Would it potentially make more sense to sort the states before you iterate through them, so that they are added to the dropdown in sorted order, rather than adding them unsorted and then sorting the dropdown? If you could get them into an array (like just the name or number), then you could use JS's built-in sort method and wouldn't have to write your own.

src/main/webapp/data-visualization.js
```diff
@@ -1,15 +1,27 @@
...
+        document.body.appendChild(script);
+        script.onload = resolve;
+        script.onerror = reject;
+        script.async = true;
+        script.src =
+        `https://www.amcharts.com/lib/4/geodata/region/usa/${abbrev}Low.js`;
+      });
+      stateInfo[location].geoJsonLoaded = true;
+    }
+    return window['am4geodata_region_usa_' + abbrev + 'Low'];
```
otobin, line 16: This is such a smart way to load the files !

src/main/webapp/data-visualization.js
```diff
@@ -283,4 +295,4 @@ function setStyle(isCountyQuery) {
     const amChartsDiv = document.getElementById('am-charts');
     amChartsDiv.style.display = 'block';
   }
-}
\ No newline at end of file
+}
```
otobin, line 298: nit: function is outdated. Should resolve itself in the master.

src/main/webapp/query-control.js
```diff
@@ -119,4 +119,48 @@ function replaceValueIfEmpty(dataListId) {
...
+    if (stateInfo.hasOwnProperty(state)) {
+      if (state !== '72') { // Skip Puerto Rico
```
otobin, line 157: Would it be better to just delete Puerto Rico from the stateInfo file?

src/main/webapp/query-control.js
otobin: ^^ You would have to iterate through the stateInfo file an extra time but having a sorted list of the numerical values of all the states might be useful for some future functions.

src/main/webapp/query-control.js
nsri-google, line 157: instead can you just remove this entry from stateinfo file.

kyracq: The Census API returns Puerto Rico as a state and I couldn't find a way to query all states excluding Puerto Rico. So, I wasn't able to remove PR from state-info without breaking the each U.S. state query.

PR #24, opened 2020-08-04
Data table

kyracq:
Updates: Use Visualization API for data table, link to census table

deployed at https://censusviz.appspot.com/
- show/hide data table corresponding to map

I think we'd also wanted to link to a table from the census such as this one https://data.census.gov/cedsci/table?q=ACS&hidePreview=true&tid=ACSDP1Y2018.DP05&vintage=2018&g=0400000US04 but I don't know of a straight forward way to get a link to the table data corresponding to our map. Let me know if you have suggestions, I was thinking just showing our own table may be sufficient


src/main/webapp/style.css
```diff
@@ -52,21 +56,32 @@ input[type=submit] {
...
   transition: all .4s ease-in-out;
 }
 
+button,
 input[type=submit]:focus {
   outline: 0;
 }
 
+button {
+  background: blue;
```
shulijones, line 75: Nit: I don't love this shade. To my eyes, it seems too bright for the relatively large size of the button. I'd prefer either something a bit darker, or to flip to having the text be black and go with a much lighter blue.

src/main/webapp/query-control.js
```diff
@@ -20,6 +20,7 @@ function registerServiceWorker() {
 // to be reformatted and visualized.
 function passQuery() {
   document.getElementById('map-title').innerText = '';
+  document.getElementById('data-table').innerHTML = '';
   am4core.disposeAllCharts();
   document.getElementById('more-info').innerText = 'Please wait. Loading...';
   document.getElementById('result').style.display = 'block';
```
shulijones, line 26: We've got a lot of stuff going on here at this point. Maybe we should make a separate smaller function, something like 'clearPreviousResult', and put all of that in there. Then it would make sense to call clearPreviousResult each time a new query is passed in.

src/main/webapp/data-visualization.js
```diff
@@ -296,3 +299,47 @@ function setStyle(isCountyQuery) {
     amChartsDiv.style.display = 'block';
   }
 }
+
+/**
+ * Create and return an HTML table using the data in dataArray.
+ * Uses first row in dataArray as table headers.
+ */
+function createDataTable(dataArray) {
```
shulijones, line 307: I sort of wonder if there's a shorter way to do this, but I can't quite think of any... you could make a createTableRow function but not sure if it really adds any efficiency.

src/main/webapp/style.css
kyracq: Toned down the blue and added dim on hover to all buttons. Let me know if you think it should be lighter/darker!

src/main/webapp/query-control.js
kyracq: Done!

src/main/webapp/data-visualization.js
kyracq: Moved the data table creation to inside the create data array function, so we don't loop through the data twice. Let me know your thoughts on this change

src/main/webapp/data-visualization.js
```diff
@@ -226,18 +232,18 @@ function getMinAndMaxPopulation(populationArray) {
 // Takes in mapsData object which has a data structure that maps
 // counties to populations, a max population, and a min population.
 // Initializes the geoJson and adds multiple event listeners.
-async function displayCountyGeoJson(mapsData, stateName) {
+async function displayCountyGeoJson(mapsData, stateNumber) {
```
otobin, line 235: good call on changing the variable name

src/main/webapp/style.css
otobin: I think it looks great! The blue theme is really nice

src/main/webapp/data-visualization.js
```diff
@@ -296,3 +302,47 @@ function setStyle(isCountyQuery) {
     amChartsDiv.style.display = 'block';
   }
 }
+
+function initDataTable(isCountyQuery) {
```
otobin, line 306: quick function description for consistency might be good but the function name makes it readable/understandable

nsri-google, line 306: This is a simple and good start. For next steps, you could also consider using google table visualization (https://developers.google.com/chart/interactive/docs/gallery/table) to display the table. It allows you sorting by any column automatically.

src/main/webapp/data-visualization.js
```diff
@@ -296,3 +302,47 @@ function setStyle(isCountyQuery) {
...
+  tableRow.appendChild(value);
+  table.appendChild(tableRow);
+  return table;
+}
+
+/**
+ * Create and return an HTML table using the data in dataArray.
+ * Uses first row in dataArray as table headers.
+ */
+function appendRowToDataTable(table, locationName, population) {
```
nsri-google, line 324: Maybe this should be combined with initDataTable. You can collect all the data in variables and have one function that creates the table and populates all the information in it.

anicr7, line 324: The data doesn't seem to be sorted by name or value. This might make it little difficult to read.

How about sorting the data either during the append or during display?

src/main/webapp/data-visualization.js
```diff
@@ -129,21 +128,28 @@ function createDataArray(censusDataArray, isCountyQuery) {
   const vizDataArray = [];
   // first row is headers
   const regionIndex = censusDataArray[0].indexOf('state');
-  censusDataArray = censusDataArray.splice(1); // get rid of header row
+  censusDataArray = censusDataArray.slice(1); // get rid of header row
   // Check to see if an extra calculation for percentages is needed
+  const table = initDataTable(isCountyQuery);
```
nsri-google, line 133: Instead of creating a table element here start collecting the data in variable(s) and then just before the 
"document.getElementById('data-table').appendChild(table);" line you can call one function that creates the table. This isolates all the table creation stuff to one function and you easily replace that with some other way to display the data should you choose to.

src/main/webapp/style.css
shulijones: Looks great, thank you!

src/main/webapp/query-control.js
```diff
@@ -14,13 +14,18 @@ function registerServiceWorker() {
...
 // Get the query the user entered and display the result.
 // Breaks down the query and passes it to the backend to be analyzed;
 // the backend returns the appropriate data, which is then passed off
 // to be reformatted and visualized.
 function passQuery() {
-  document.getElementById('map-title').innerText = '';
-  am4core.disposeAllCharts();
+  clearPreviousResult();
   document.getElementById('more-info').innerText = 'Please wait. Loading...';
   document.getElementById('result').style.display = 'block';
```
shulijones, line 30: Could these also go into clearPreviousResult? Not sure, open to your thoughts

src/main/webapp/data-visualization.js
```diff
@@ -296,3 +302,47 @@ function setStyle(isCountyQuery) {
...
+ */
+function appendRowToDataTable(table, locationName, population) {
+  const tableRow = document.createElement('tr');
+  const region = document.createElement('td');
+  const value = document.createElement('td');
+  region.innerText = locationName;
+  value.innerText = population;
+  tableRow.appendChild(region);
+  tableRow.appendChild(value);
+  table.appendChild(tableRow);
```
shulijones, line 332: Some of this looks very similar to initDataTable; is there a way to write those lines of code just once?

src/main/webapp/data-visualization.js
shulijones: I think that's a great idea! I see there are some other comments on that section now, but overall the move makes sense to me.

src/main/java/QueryServlet.java
```diff
@@ -15,39 +16,94 @@
 @WebServlet("/query")
 public class QueryServlet extends HttpServlet {
 
-  // Hardcoded initial choices for which lines of the data table are accessed
+  // Hardcoded initial choices for which lines of which data table are accessed
   Map<String, Map<String, String>> queryToDataRow =
       ImmutableMap.of(
           "live",
-          ImmutableMap.of("under-18", "0019E", "over-18", "0021E", "all-ages", "0001E"),
+          ImmutableMap.of(
```
shulijones, line 23: I know the linting recommended all these line breaks, but it doesn't seem to be recommended in the Java style guide (or could you point me to where it is?) It seems sort of inefficient to me as a formatting strategy. Happy to hear arguments otherwise however

src/main/webapp/data-visualization.js
```diff
@@ -306,38 +301,24 @@ function setStyle(isCountyQuery) {
...
-  tableRow.appendChild(value);
-  table.appendChild(tableRow);
+// Draw data table using Visualization API
+function drawTable(dataArray, isCountyQuery) {
+  google.charts.load('current', {'packages': ['table']});
+  google.charts.setOnLoadCallback(() => {
+    const data = new google.visualization.DataTable();
+    const nameHeader = isCountyQuery ? 'County' : 'State';
+    data.addColumn('string', nameHeader);
+    data.addColumn('number', 'Population');
```
shulijones, line 311: We probably want to change this text to be something more dynamic, since it's not always Population, but that can be a later job

src/main/webapp/data-visualization.js
```diff
@@ -306,38 +301,24 @@ function setStyle(isCountyQuery) {
...
+    dataArray.forEach((elem) => {
+      data.addRow([elem.name, parseInt(elem.value)]);
+    });
+    const table = new google.visualization.Table(
+        document.getElementById('data-table'));
+    table.draw(data, {width: '100%', height: '100%'});
```
shulijones, line 317: Do we want to reduce the width of the table? It stretches across my screen but it's mostly empty space, and the numbers and the text end up really far away from each other. I think it might read more easily at approximately the width you had before

src/main/webapp/query-control.js
```diff
@@ -56,23 +57,27 @@ function passQuery() {
...
-          .then((censusDataArray) => {
+        response.json()
+        .then((response) => {
             // censusDataArray is a 2D array, where the first row is a
             // header row and all subsequent rows are one piece of
             // data (e.g. for a state or county)
-            displayVisualization(censusDataArray, description,
+            displayVisualization(JSON.parse(response.data), description,
               location, isCountyQuery);
+            displayLinkToCensusTable(response.table, year);
```
shulijones, line 80: I like the way you set this up, so that the info goes from the backend, to the control, and then directly to a new JS function (rather than getting dragged along through all our other JS functions).

src/main/webapp/data-visualization.js
```diff
@@ -349,3 +330,13 @@ function toggleDataTable() {
     document.getElementById('toggle-data-btn').innerText = 'Display raw data';
   }
 }
+
+// Display link to data.census.gov table for the table the displayed
+// data is from.
+function displayLinkToCensusTable(tableString, year) {
+  const tablePrefix = tableString[0] === 'S' ? 'SPP' : 'DP';
```
shulijones, line 337: There's actually three kinds of tables: the ones that start with DP, the ones that start with SPP, and the ones that start with S (unfortunately). It looks like this doesn't take into account the ones that start with S.
Also, those are just the three kinds we're using right now, so potentially we might have to change this code if we added a fourth kind. Is there a way to do these calculations in the Java backend, so that the javascript doesn't have to know anything about SPP vs DP vs S, and it's all kept in one place when we need to change it?

src/main/webapp/data-visualization.js
otobin: This may be a hot take but I think I actually like it better without the data table API. Maybe we could adjust the width/font size of this table to make it look nicer but I'm still not a fan of the header in this one. Happy to hear any other thoughts though

shulijones: I agree that the API makes it look super weird. But maybe there is room for adjustment? Like to make the header not have that weird gradient?

src/main/webapp/data-visualization.js
kyracq: Thanks for letting me know, moved link generation all to backend

src/main/webapp/data-visualization.js
kyracq: The visualization API allows configuring the table with any styles we'd like, for now I just removed the header gradient. Also may the width 30%

src/main/java/QueryServlet.java
kyracq: I agree this seems kinda inefficient, but I do think it looks cleaner, and as we add more query dropdowns these maps will only get longer and we'd have to break the maps into multiple lines anyway. I don't see this mentioned in the written style guide however the linter I added is straight from google so I'm guessing anything the linter suggests aligns with the style guide

shulijones: OK, makes sense!

src/main/webapp/data-visualization.js
otobin: It would be pretty easy to pass in description in this function from displayVisualization

PR #25, opened 2020-08-05
Add man and woman queries; update percentage checker

shulijones:
Add the option to ask about "men" or "women" via hardcoded queries. Also, change the percentage checker so that it is looking at numbers, not at hardcoded headers.

Note that actually, there was a bug in the percentage checker - it was originally using the data's header row to check the query name against a hardcoded list, but we started removing the header row earlier, so it was actually looking at the first row of real data.

### DIFF
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -37,7 +37,7 @@ public class QueryServlet extends HttpServlet {
     } else if (tablePrefix.substring(0, 1).equals("S")) {
       return "subject?get=NAME,";
     }
-    return ""; // Will throw error when query is made; should never reach this point
+    return ""; // should never reach this point
   }
 
   @Override
@@ -61,10 +61,16 @@ public class QueryServlet extends HttpServlet {
     }
 
     String dataRow = queryToDataRow.get(action).get(personType);
+    String dataTablePrefix = getDataTableString(dataRow);
+    if (dataTablePrefix.equals("")) {
+      response.sendError(
+          HttpServletResponse.SC_NOT_IMPLEMENTED, "We do not support this visualization yet.");
+    }
+
     URL fetchUrl =
         new URL(
             "https://api.census.gov/data/2018/acs/acs1/"
-                + getDataTableString(dataRow)
+                + dataTablePrefix
                 + dataRow
                 + "&for="
                 + (location.equals("state") ? "state:*" : "county:*&in=state:" + location)
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -46,15 +46,21 @@ public class QueryServlet extends HttpServlet {
     String action = request.getParameter("action");
     String location = request.getParameter("location");
 
-    if (!queryToDataRow.containsKey(action)
-        || !queryToDataRow.get(action).containsKey(personType)) {
-      // We don't have a data table for this query
+    if (!queryToDataRow.containsKey(action)) {
+      // We don't have information on this action
       response.sendError(
-          HttpServletResponse.SC_BAD_REQUEST, "We do not support this visualization yet.");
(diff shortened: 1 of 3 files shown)


src/main/webapp/data-visualization.js
```diff
@@ -164,20 +164,17 @@ function percentToTotal(totalNumber, percentage) {
...
-      return true;
-    }
+  // percentageQueries is a list of queries that return percents and not raw data.
+  // These queries have two columns of numbers. One is a total number and one
+  // is a number between 0 and 100 (representing the percentage of the total)
+  if (headerColumn.length !== 4) {
+    return false;
   }
-  return false;
+  firstNum = Number(headerColumn[1]);
```
kyracq, line 173: missing declaration/const

shulijones: Nice catch, thanks.

src/main/java/QueryServlet.java
```diff
@@ -15,35 +15,57 @@
...
       return;
     }
 
+    String dataRow = queryToDataRow.get(action).get(personType);
     URL fetchUrl =
         new URL(
             "https://api.census.gov/data/2018/acs/acs1/"
-                + (action.equals("live") ? "profile?get=NAME,DP05_" : "spp?get=NAME,S0201_")
-                + queryToDataRow.get(action).get(personType)
+                + getDataTableString(dataRow)
```
anicr7, line 67: We shouldn't do a query when this returns "". We can respond back with an error instead.

src/main/webapp/data-visualization.js
```diff
@@ -131,13 +131,13 @@ function createDataArray(censusDataArray, isCountyQuery) {
   censusDataArray = censusDataArray.slice(1); // get rid of header row
```
otobin, line 131: Maybe I'm looking at an outdated version, but I think the bug might still be here -- when you reassign censusDataArray to censusDataArray.slice(1), you get rid of the header, so the reference to censusDataArray[0] in line 134 will actually be censusDataArray[1]. An adjustment that you could make is define another variable, censusArrayWithoutHeader = censusDataArray.slice(1) and then traverse that, so that when you reference censusDataArray later it is still the original data.

src/main/webapp/data-visualization.js
```diff
@@ -156,28 +156,26 @@ function createDataArray(censusDataArray, isCountyQuery) {
...
-  // Iterate through all the data variables in the header.
-  // Eg: ["NAME","S0201_119E","S0201_126E","state"]
-  // We need to return true
-  for (i = 1; i < headerColumn.length - 1; i++) {
-    // if the current data is in the percentage array
-    if (percentageQueries.indexOf(headerColumn[i]) > -1) {
-      return true;
-    }
+function checkPercentage(headerColumn, isCountyQuery) {
+  // percentageQueries is a list of queries that return percents and not raw data.
```
otobin, line 166: I think you can delete these comments bc they refer to an earlier version of this function

otobin, line 166: percentageQueries is no longer a variable in this function so it might be helpful to update the comment with only explanations for why percentage queries have length 5 instead of 4

src/main/webapp/data-visualization.js
shulijones: Yeah, you're definitely right that the header is still getting erased. But I realized that actually, it makes a lot of sense to look at the first row and not the header row. The header can't tell us what's a percentage and what's not, but the numbers themselves can.

src/main/webapp/data-visualization.js
shulijones: Nice catch, thanks.

src/main/java/QueryServlet.java
shulijones: Done.

PR #28, opened 2020-08-05
All states geo json

kyracq:


src/main/webapp/data-visualization.js
```diff
@@ -133,21 +132,28 @@ function createDataArray(censusDataArray, isCountyQuery) {
   const vizDataArray = [];
   // first row is headers
   const regionIndex = censusDataArray[0].indexOf('state');
-  censusDataArray = censusDataArray.splice(1); // get rid of header row
+  censusDataArray = censusDataArray.slice(1); // get rid of header row
   // Check to see if an extra calculation for percentages is needed
+  const table = initDataTable(isCountyQuery);
   if (checkPercentage(censusDataArray[0])) {
```
shulijones, line 138: I updated this a bit (checkPercentage now also takes isCountyQuery as a parameter) but I guess that'll get merged when we merge data-table

src/main/webapp/data-visualization.js
```diff
@@ -299,3 +305,47 @@ function setStyle(isCountyQuery) {
   const mapsDiv = document.getElementById('map');
   mapsDiv.style.display = 'none';
 }
+
+function initDataTable(isCountyQuery) {
```
shulijones, line 309: I left some comments on the table creation functions in data-table but I'm assuming that we can keep that over there and just merge carefully later

PR #29, opened 2020-08-06
Colors

otobin:
User is now able to change the color of both visualizations from a color input. 

Because the Google Maps API prevents the direct caching of maps, I cached the color and other necessary data produced from the fetch in localStorage and called the displayAmCharts() and displayGeoJson() again with the necessary parameters. This also allows colors to stay rather than resetting after each query.


src/main/webapp/query-control.js
```diff
@@ -17,6 +17,8 @@ function registerServiceWorker() {
 function clearPreviousResult() {
   document.getElementById('map-title').innerText = '';
   document.getElementById('data-table').innerHTML = '';
+  document.getElementById('colors').style.display = 'none';
+  document.getElementById('setColor').value = "#0000ff";
```
shulijones, line 21: How about #0071bd instead? That's the color that the buttons are

src/main/webapp/style.css
```diff
@@ -129,3 +129,7 @@ footer {
   text-align: center;
   width: 100%;
 }
+
+#colors {
+  text-align: center;
+}
```
shulijones, line 135: Taking a look through this css, we have text-align: center on almost everything. Would it be possible to just apply it to one larger thing (e.g. the root) instead? (I'm not sure if this will work or not, so might have to experiment.)

src/main/webapp/index.html
```diff
@@ -85,6 +85,10 @@ <h1>Census Viz</h1>
     <div id="result">
       <div id="map-title"></div>
       <div id="more-info"></div>
+      <div id="colors">
+        <input type="color" id="setColor" name="setColor" value="#0000ff" onchange="changeColor(this)">
+        <label for="head">Change the color of the visualization</label>
```
shulijones, line 90: I think typically, the label should go before the input, and the label's for attribute should be equal to the id of the input element (so in this case, setColor).

src/main/webapp/data-visualization.js
```diff
@@ -294,13 +301,28 @@ function setStyle(isCountyQuery) {
...
-
   const chartsDiv = document.getElementById('am-charts');
   chartsDiv.style.display = 'block';
   const mapsDiv = document.getElementById('map');
   mapsDiv.style.display = 'none';
 }
 
+// Changes the color of the current visualizations on the page. 
+function changeColor(colorParam) {
+  if (typeof map !== 'undefined') {
```
shulijones, line 312: In what situation would map be undefined? If it is, should we be throwing an error or something?

src/main/webapp/data-visualization.js
```diff
@@ -294,13 +301,28 @@ function setStyle(isCountyQuery) {
...
+      mapData.maxValue]);
+    map.data.forEach(function(feature) {
+      map.data.setStyle((feature) => {
+        return {
+          fillColor: colorScale(mapData.map[feature.j.name]).toString(),
+        };
+      });
+    });
+  }
+  displayAmChartsMap(amChartsData, globalDescription, globalGeoData, colorParam.value);
```
shulijones, line 323: I think I see why you needed to use global variables, so that you could just call the displayAmChartsMap function again without re-doing every single aspect of the visualization. 
Here's an idea: what if we did basically this, but instead of having the global variables actually define what gets visualized (which seems risky), we make them be copies? So I'm essentially imagining that we have a global cache. The cache stores only one set of objects, the ones that were most recently created in displayVisualization. Then, we could have a displayCachedAmChartsMap function, which gets the variables out of the cache and sends them to  displayAmChartsMap. 
That way, inside displayVisualization, we're not working with global variables, we can just use local ones. But when everything is settled, we save it separately globally to be re-accessed if necessary.
This is just one approach so definitely interested to hear Kyra's thoughts as well.

src/main/webapp/query-control.js
otobin: Fixed

src/main/webapp/index.html
otobin: fixed!

src/main/webapp/data-visualization.js
otobin: I haven't implemented the geoJson for state queries, so for state queries maps is undefined. Added comment to clarify this.

src/main/webapp/index.html
```diff
@@ -85,6 +85,10 @@ <h1>Census Viz</h1>
     <div id="result">
       <div id="map-title"></div>
       <div id="more-info"></div>
+      <div id="colors">
+        <input type="color" id="setColor" name="setColor" value="#0000ff" onchange="changeColor(this)">
```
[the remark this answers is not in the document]
otobin: nice catch! fixed

src/main/webapp/data-visualization.js
```diff
@@ -294,21 +300,38 @@ function setStyle(isCountyQuery) {
...
+  const color = colorParam.value;
+  localStorage.setItem('color', color);
+  // Get variables out of cache
+  const cacheMapsData = JSON.parse(localStorage.getItem('mapsData'));
+  const cacheGeoData = JSON.parse(localStorage.getItem('geoData'));
+  const cacheAmCharts = JSON.parse(localStorage.getItem('amChartsData'));
+  const cacheLocation = localStorage.getItem('location');
+  const cacheDescription = localStorage.getItem('description');
+  // map is undefined on a state query, so check to be sure that 
+  // it is undefined before calling displayCountyGeoJson.
```
shulijones, line 320: nit: check to be sure that it isn't undefined

src/main/webapp/data-visualization.js
otobin: Changed to caching  all of the necessary data in localStorage. The google Maps api does not support directly caching maps.

src/main/webapp/data-visualization.js
otobin: updated.

src/main/webapp/data-visualization.js
```diff
@@ -228,38 +231,41 @@ function getMinAndMaxPopulation(populationArray) {
...
+  const colorScale = chroma.scale([minColor, maxColor]).domain([minPopulation,
     maxPopulation]);
 
-  const geoData = await getGeoData(stateNumber, true);
   map.data.addGeoJson(geoData);
   map.data.forEach(function(feature) {
     map.data.setStyle((feature) => {
       return {
         fillColor: colorScale(countyToPopMap[feature.j.name]).toString(),
+        fillOpacity: 0.4,
```
shulijones, line 257: I'd be interested in also seeing what 0.5 and 0.6 looked like - have you tried them out?

src/main/webapp/data-visualization.js
shulijones: it looks the same to me? (sorry!)

src/main/webapp/query-control.js
```diff
@@ -14,9 +14,20 @@ function registerServiceWorker() {
   }
 }
 
+// Sets the value of the color input. Checks to see if a color is in 
+// the cache already, and returns a nice blue if it isn't.
+function getColor() {
+  if (localStorage.getItem('color') !== null) {
+    document.getElementById('setColor').value = localStorage.getItem('color');
+  } else {
+    document.getElementById('setColor').value = '#0071bd';
```
kyracq, line 23: localStorage color could be given initial value here

src/main/webapp/index.html
```diff
@@ -86,16 +94,20 @@ <h1>Census Viz</h1>
       </div>
       <div id="map-title"></div>
       <div id="more-info"></div>
+      <div id="colors">
+        <label for="setColor">Change the color of the visualization</label>
+        <input type="color" id="setColor" name="setColor" onchange="changeColor(this)">
```
kyracq, line 99: I think you'd changed the id already but it's not here anymore for some reason

src/main/webapp/data-visualization.js
```diff
@@ -312,16 +318,34 @@ function setStyle(isCountyQuery) {
...
+  localStorage.setItem('color', color);
+  // Get variables out of cache
+  const cacheMapsData = JSON.parse(localStorage.getItem('mapsData'));
+  const cacheGeoData = JSON.parse(localStorage.getItem('geoData'));
+  const cacheAmCharts = JSON.parse(localStorage.getItem('amChartsData'));
+  const cacheLocation = localStorage.getItem('location');
+  const cacheDescription = localStorage.getItem('description');
+  // map is undefined on a state query, so check to be sure that 
+  // it is undefined before calling displayCountyGeoJson.
+  if (typeof cacheMapsData !== 'undefined') {
```
nsri-google, line 333: instead you should call a function that is called both here and in displayVisualization. Here you get values from local storage and then call the function whereas in displayVisualization you retreive values from backend, store them in local storage and call the same function.

src/main/webapp/query-control.js
```diff
@@ -14,9 +14,20 @@ function registerServiceWorker() {
   }
 }
 
+// Sets the value of the color input. Checks to see if a color is in 
+// the cache already, and returns a nice blue if it isn't.
+function getColor() {
+  if (localStorage.getItem('color') !== null) {
```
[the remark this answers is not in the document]
otobin: fixed.

src/main/webapp/query-control.js
otobin: true! fixed

src/main/webapp/query-control.js
```diff
@@ -14,9 +14,20 @@ function registerServiceWorker() {
   }
 }
 
+// Sets the value of the color input. Checks to see if a color is in
```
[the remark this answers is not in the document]
otobin: Fixed function name to match id and comment.

src/main/webapp/data-visualization.js
otobin: Should be fixed now but it isn't showing up on my computer so let me know if there are any more bugs!

src/main/webapp/data-visualization.js
otobin: Changed to .5. At .6 it becomes a little difficult to see text for cities inside the counties.

src/main/webapp/index.html
otobin: I did revert back to a previous commit because of a bug at some point. Either way, it should be working now. Let me know if there are any other issues.

src/main/webapp/data-visualization.js
```diff
@@ -21,26 +21,34 @@ async function getGeoData(location, isCountyQuery) {
   }
 }
 
-// Display amCharts and geoJson visulizations and create data table
-// for given data.
+// Display amCharts and geoJson visulizations for given data.
 async function displayVisualization(censusDataArray, description,
   location, isCountyQuery) {
+  document.getElementById('colors').style.display = 'block';
+  getColor();
```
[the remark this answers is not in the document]
otobin: getColor (now named setColor) also adds the default blue color to the cache, so I think I will leave as is.

src/main/webapp/data-visualization.js
otobin: Do you mean one function to put values in the cache and one function to fetch values from the cache? I understand where I'm supposed to call the function but I'm not really sure what the function does

src/main/webapp/query-control.js
```diff
@@ -16,11 +16,13 @@ function registerServiceWorker() {
 
 // Sets the value of the color input. Checks to see if a color is in 
 // the cache already, and returns a nice blue if it isn't.
-function getColor() {
-  if (localStorage.getItem('color') !== null) {
+function setColor() {
+  const color = localStorage.getItem('color');
+  if (color !== null) {
     document.getElementById('setColor').value = localStorage.getItem('color');
```
nsri-google, line 22: use the color variable here

src/main/webapp/data-visualization.js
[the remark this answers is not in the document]
nsri-google: you can avoid yet another call to localStorage.getItem if you just return the value from setColor().

so instead of setColor();
const color = localStorage.getItem('color');

you can simply have const color = setColor();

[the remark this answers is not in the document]
nsri-google: your setColor function can be like this.

function setColor() {
  const color = localStorage.getItem('color') || '#0071bd';
  document.getElementById('setColor').value = color;
  return color;
}

src/main/webapp/data-visualization.js
nsri-google, line 333: This would only be undefined at the very beginning when you are showing the full US map and so not saving this value to localStorage. But once you make a county query that value will be stored here and since you never delete this value it will have the last save values. so maybe you have actually have to pass in whether this is a county query or not as a parameter and make this a conditional on that.

src/main/webapp/data-visualization.js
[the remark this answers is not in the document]
otobin: I see, this definitely alleviates the amount of calls to localStorage. Fixed.

src/main/webapp/query-control.js
otobin: nice catch. fixed.

src/main/webapp/data-visualization.js
```diff
@@ -25,8 +25,8 @@ async function getGeoData(location, isCountyQuery) {
 async function displayVisualization(censusDataArray, description,
   location, isCountyQuery) {
   document.getElementById('colors').style.display = 'block';
-  setColor();
-  const color = localStorage.getItem('color');
+  const color = setColor();
+  document.getElementById('setColor').value = color;
```
nsri-google, line 29: Since you now moved the 'set' part here the function should actually be called getColor now.

src/main/webapp/data-visualization.js
otobin: location is actually "state" for a state query and "01", "02" etc... for a county query where the location represents the state number, so I updated the conditional based on when cacheLocation is not "state"

src/main/webapp/data-visualization.js
otobin: true! fixed.

src/main/webapp/data-visualization.js
```diff
@@ -25,12 +25,13 @@ async function getGeoData(location, isCountyQuery) {
...
   location, isCountyQuery) {
   document.getElementById('colors').style.display = 'block';
-  const color = setColor();
+  const color = getColor();
   document.getElementById('setColor').value = color;
   const geoData = await getGeoData(location, isCountyQuery);
   // Put all necessary data in the cache
   localStorage.setItem('geoData', JSON.stringify(geoData));
   localStorage.setItem('location', location);
+  console.log(location);
```
nsri-google: remove this line

PR #30, opened 2020-08-06
Add toggle switch with understandable names

kyracq:
Not currently deployed

Screenshot 2020-08-06 at 4 55 25 PM
Screenshot 2020-08-06 at 4 55 35 PM

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -76,14 +76,13 @@
     </div>
     <div id="result">
       <div id="map-options">
-        <div class="map-option">Flat view</div>
         <div class="map-option">
           <label class="switch" id="map-toggle-switch">
             <input type="checkbox" id="map-toggle" onchange="toggleMap()">
             <span class="slider round"></span>
           </label>
         </div>
-        <div class="map-option">Map overlay</div>
+        <div class="map-option" id="map-toggle-msg">Enable map overlay</div>
       </div>
       <div id="map-title"></div>
       <div id="more-info"></div>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -26,56 +26,48 @@
       <form id="query-form" onsubmit="passQuery(); return false;">
         <p>
           How many
-          <input type="list" list="person-type" name="person-type" id="person-type-list"
-              autocomplete="off" required onkeyup="validateInput('person-type')"
-              onfocus="storeValueAndEmpty('person-type')"
-              onfocusout="replaceValueIfEmpty('person-type')"
-              value="people" pattern="children|adults|people|men|women">
-            <datalist id="person-type">
-              <option data-value="under-18" value="children"></option>
-              <option data-value="over-18" value="adults"></option>
-              <option data-value="all-ages" value="people"></option>
-              <option data-value="male" value="men"></option>
-              <option data-value="female" value="women"></option>
-            </datalist>
+          <input type="list" list="person-type" name="person-type" id="person-type-list" autocomplete="off" required
+            onkeyup="validateInput('person-type')" onfocus="storeValueAndEmpty('person-type')"
+            onfocusout="replaceValueIfEmpty('person-type')" value="people" pattern="children|adults|people|men|women">
+          <datalist id="person-type">
+            <option data-value="under-18" value="children"></option>
+            <option data-value="over-18" value="adults"></option>
+            <option data-value="all-ages" value="people"></option>
+            <option data-value="male" value="men"></option>
+            <option data-value="female" value="women"></option>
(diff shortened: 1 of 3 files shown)


src/main/webapp/style.css
```diff
@@ -98,6 +98,78 @@ input[type=submit]:hover {
   text-align: center;
 }
 
+/* The switch - the box around the slider */
+.switch {
+  position: relative;
+  display: inline-block;
```
shulijones, line 104: When I ask an all-states query, the first thing that happens is that the title appears, the toggle button appears, and the loading text appears. The second thing that happens is that the map replaces the loading text and the toggle button disappears. If the loading takes a few seconds, then it's really noticeable that the toggle button appears only to disappear again.
To avoid this, maybe we should have the toggle be default hidden, and it only gets made visible if it's a specific-state query, rather than vice versa.

src/main/webapp/index.html
```diff
@@ -26,69 +26,67 @@ <h1>Census Viz</h1>
...
+          </datalist>
           </input>
           ?
         </p>
         <input type="submit" value="go">
       </form>
     </div>
     <div id="result">
+      <div id="map-options">
+        <div class="map-option">Flat map</div>
```
shulijones, line 79: The name "Flat map" makes a lot of sense to me. But then "map overlay" doesn't make as much sense, because I don't know what's being overlaid (the word "map" is already being used to mean the other thing).
What about "Flat view" vs "Map overlay" or "Map view"? That way we save the word map for just the geoJson.
Or, what about "Flat map" vs "Terrain"? Terrain isn't exactly right to describe what the geoJson shows, but it's pretty close.

src/main/webapp/index.html
```diff
@@ -26,69 +26,67 @@ <h1>Census Viz</h1>
...
+        <div class="map-option">
+          <label class="switch" id="map-toggle-switch">
+            <input type="checkbox" id="map-toggle" onchange="toggleMap()">
+            <span class="slider round"></span>
+          </label>
+        </div>
+        <div class="map-option">Map overlay</div>
+      </div>
       <div id="map-title"></div>
```
shulijones, line 88: Could we add a line break here potentially, or alternatively put some padding around the toggle using CSS? The toggle, the title, and the map are really crowded together, and I think it should be more clear that the title goes with the map. (That will also let people screenshot those two parts alone more easily if they want to, without getting the toggle in the way -- I'm always annoyed when a site makes it hard for me to do that.)

shulijones: Oh, this will be even more important when colors and map toggle are merged. I'm doing that now for my PR and it's getting kind of messy up there lol.

kyracq: Good point, added some space between toggle switch and map title

src/main/webapp/index.html
kyracq: Changed to "Flat view" and "Map overlay"

src/main/webapp/style.css
kyracq: Made toggle hidden by default!

src/main/webapp/style.css
```diff
@@ -93,9 +93,78 @@ input[type=submit]:hover {
...
+.slider {
+  position: absolute;
+  border-radius: 34px;
+  cursor: pointer;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: #ccc;
+  -webkit-transition: .4s;
```
otobin, line 121: This is so nitpicky so feel free to disregard but right now the div's visibility changes instantaneously even though toggling the map takes .4 seconds. It would be cool to create some fade in animations that correspond to the toggle for each div, but this is obviously pretty low priority

src/main/webapp/index.html
otobin: Late but I like these names

src/main/webapp/style.css
kyracq: Yeah a fade in would look nice, will look into doing this

src/main/webapp/index.html
```diff
@@ -26,69 +26,67 @@ <h1>Census Viz</h1>
...
+          </datalist>
           </input>
           ?
         </p>
         <input type="submit" value="go">
       </form>
     </div>
     <div id="result">
+      <div id="map-options">
+        <div class="map-option">Flat view</div>
```
nsri-google, line 79: How about making this an "Enable Map Overlay" checkbox which means you do not have to call the "Flat View" anything. This checkbox will enable (or disable) the map overlay

kyracq: Thanks for this suggestion, switched to this

PR #31, opened 2020-08-06
Improve text

shulijones:
Improve text quality across the site, including in the map title, when mousing over geographical areas, and in the data table header.

All text should now make grammatical sense and change dynamically based on the query asked. There is still some hardcoding, but less than previously, and it's reused where possible (i.e. rather than three different spots hardcoding separately, doing it in one spot and then referring back to that variable).


src/main/webapp/data-visualization.js
```diff
@@ -33,11 +33,11 @@ async function displayVisualization(censusDataArray, description,
...
   globalDescription = description;
-  drawTable(amChartsData, isCountyQuery);
+  drawTable(amChartsData, globalDescription, isCountyQuery);
   if (isCountyQuery) {
       const mapsData = getMapsData(censusDataArray);
       displayAmChartsMap(amChartsData, globalDescription, globalGeoData, '#3c5bdc');
-      displayCountyGeoJson(mapsData, location);
+      displayCountyGeoJson(mapsData, globalDescription, location);
   } else {
       displayAmChartsMap(amChartsData, description, globalGeoData, '#3c5bdc');
```
kyracq, line 42: 'description' could be 'globalDescription' on this line for consistency throughout the function

src/main/java/QueryServlet.java
```diff
@@ -105,12 +105,16 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    String censusTableLink = 
+    String censusTableLink =
         "https://data.census.gov/cedsci/table?tid=ACS"
-            + tableNameToAbbrev.get(fetchUrlString.substring(
-              fetchUrlString.indexOf("acs1/") + 5, fetchUrlString.indexOf("?")))
-            + "1Y" + year + "."
-            + fetchUrlString.substring(fetchUrlString.indexOf(",") + 1, fetchUrlString.indexOf("_"));
+            + tableNameToAbbrev.get(
+                fetchUrlString.substring(
+                    fetchUrlString.indexOf("acs1/") + 5, fetchUrlString.indexOf("?")))
```
otobin, line 112: Isn't fetchUrlString.indexOf("acs1/") + 5 just the data table variable? Or no?

src/main/java/QueryServlet.java
```diff
@@ -105,12 +105,16 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
     String fetchUrlString = fetchUrl.toString();
-    String censusTableLink = 
+    String censusTableLink =
         "https://data.census.gov/cedsci/table?tid=ACS"
-            + tableNameToAbbrev.get(fetchUrlString.substring(
-              fetchUrlString.indexOf("acs1/") + 5, fetchUrlString.indexOf("?")))
-            + "1Y" + year + "."
-            + fetchUrlString.substring(fetchUrlString.indexOf(",") + 1, fetchUrlString.indexOf("_"));
+            + tableNameToAbbrev.get(
```
otobin, line 110: Might be also useful to break these lines up into multiple functions like getCensusTableLink and getFetchUrlString

src/main/java/QueryServlet.java
shulijones: Oh, I think this is actually code that Kyra wrote for the data table. @kyracq ?

src/main/java/QueryServlet.java
shulijones: @kyracq

src/main/webapp/query-control.js
```diff
@@ -45,23 +45,22 @@ function passQuery() {
...
-  const description = {action: action, age: personType};
+  const actionToPerson = new Map();
+  actionToPerson.set(
+        'live', 'Population',
+      ).set(
+        'work', 'Workers',
+      ).set(
+        'moved', 'New inhabitants',
+      );
+  const description = `${actionToPerson.get(action)} (${personType.replace('-', ' ')})`;
```
kyracq, line 56: For some reason I don't love the current look with the parentheses, I think I'd prefer it to be like "personType action" ie. Male Workers, New Male Inhabitants, but this would probably not translate well for some queries. Open to thoughts

shulijones: Yeah, I thought about this a lot. Male workers etc. would be way better, but then it doesn't work for age... Do you think it's worth hardcoding in two different description types based on the personType?

src/main/webapp/data-visualization.js
shulijones: Thanks! Fixed

src/main/java/QueryServlet.java
kyracq: Yes, thanks for catching this will change this to indexOf(dataTablePrefix)

src/main/java/QueryServlet.java
kyracq: Good point, will do this!

src/main/webapp/query-control.js
kyracq: Yeah I guess the way it is now works better since it's more dynamic. Switching it would overly complicate things, and the descriptions are currently clear to a user

PR #32, opened 2020-08-07
Fix Puerto Rico and negative values bug, minor changes in QueryServelet

kyracq:


src/main/webapp/query-control.js
```diff
@@ -86,6 +87,20 @@ function passQuery() {
     });
 }
 
+// Remove incorrect data returned by the census API
+// such as Puerto Rico data and negative numbers
+function removeErroneousData(dataArray) {
+  dataArray.forEach((elem, index) => {
+    // Using splice, directly modifies array
+    elem.forEach((item) => {
+      if (item === null || item < 0 || item > 400000000) {
```
shulijones, line 96: Maybe add comment on why you're splicing out items > 400,000,000? Not sure if necessary but might be useful documentation (slash we can pretend our code will be useful in 30 years when the population is much higher)

PR #33, opened 2020-08-07
Update API queries to access more counties' data

shulijones:
Update backend's fetch URL creation to access data tables with information for more counties than previously.

The data tables being accessed cover all counties with population > 20,000, whereas the previous tables covered all counties > 65,000. However, these tables don't exist for all queries, and they only exist after 2013. For some queries, and all queries pre-2013, you can expect to see the same behavior as before. For a good number of queries post-2013, you should see many more (but not all) counties with data.

### DIFF
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -97,7 +97,7 @@ public class QueryServlet extends HttpServlet {
             : (tableNameToAbbrev.get(dataTablePrefix.substring(1)) + "1Y"))
         + year
         + "."
-        + dataRow.substring(0, dataRow.indexOf("_"));
+        + dataRow.substring(0, (dataRow.contains("_") ? dataRow.indexOf("_") : dataRow.length() + 1));
   }
 
   @Override
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -4,7 +4,6 @@ import com.google.gson.JsonObject;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.lang.NoSuchFieldException;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.Map;
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -4,6 +4,7 @@ import com.google.gson.JsonObject;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.lang.NoSuchFieldException;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/main/java/QueryServlet.java
```diff
@@ -50,45 +51,68 @@
...
+      return "/spp"; // Special case - different from the other S tables
+    } else if (firstChar.equals("S")) {
+      return "/subject";
     }
-    return ""; // should never reach this point
+    return "N/A"; // should never reach this point
   }
 
-  private String getcensusTableLink(String fetchUrlString, String dataTablePrefix, String year) {
+  private String getcensusTableLink(String dataRow, String dataTablePrefix, String year) {
```
anicr7, line 92: Change this to getCensusTableLink

src/main/java/QueryServlet.java
```diff
@@ -50,45 +51,68 @@
...
   // Depending on the beginning of the data table string, the query URL changes slightly
   private String getDataTableString(String tablePrefix) {
-    if (tablePrefix.substring(0, 1).equals("D")) {
-      return "profile?get=NAME,";
+    String firstChar = tablePrefix.substring(0, 1);
+    if (firstChar.equals("K")) {
+      return "";
+    } else if (firstChar.equals("D")) {
+      return "/profile";
     } else if (tablePrefix.substring(0, 5).equals("S0201")) {
```
anicr7, line 84: Substring calls with endPrefix > length of string will cause an IndexOutOfBoundsException to be thrown. So we should be careful with these calls as it can cause the Servlet to crash.

src/main/java/QueryServlet.java
```diff
@@ -50,45 +51,68 @@
...
-      return "spp?get=NAME,"; // Special case - different from the other S tables
-    } else if (tablePrefix.substring(0, 1).equals("S")) {
-      return "subject?get=NAME,";
+      return "/spp"; // Special case - different from the other S tables
+    } else if (firstChar.equals("S")) {
+      return "/subject";
     }
-    return ""; // should never reach this point
+    return "N/A"; // should never reach this point
```
anicr7, line 89: Not a fan of magic words like this. Even empty was risky, the best way to do this would be to return an exception and catch it instead.

For now you can ignore this.

src/main/java/QueryServlet.java
```diff
@@ -97,9 +121,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       return;
     }
 
-    String dataRow = queryToDataRow.get(action).get(personType);
+    String dataRow;
```
anicr7, line 124: How about 

String dataRow = queryToDataRowGeneric.get(action).get(personType);
if (year > 2013
        && queryToDataRowPost2013.containsKey(action)
        && queryToDataRowPost2013.get(action).containsKey(personType)) {
      dataRow = queryToDataRowPost2013.get(action).get(personType);
}

Then you don't need the else and its easier to read.

src/main/java/QueryServlet.java
shulijones: You're right, I felt bad just writing that. I will throw an exception

src/main/java/QueryServlet.java
shulijones: That's a good point. The input will definitely always have at least one character, but five is not as much of a guarantee so I will add an extra check there.

src/main/java/QueryServlet.java
shulijones: Good idea, thank you! Done.

PR #34, opened 2020-08-10
Handle no data

kyracq:

Added:
- Favicon
- Data not available message for counties with no data


src/main/java/QueryServlet.java
```diff
@@ -111,15 +117,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      response.sendError(
-          HttpServletResponse.SC_BAD_REQUEST,
-          "This query is not supported by census data. Try asking a more general one.");
+      response.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
+      response.setContentType("application/json;");
+      response
+          .getWriter()
+          .println(
+              sendError(
+                  "This query is not supported by census data. Try asking a more general one."));
```
shulijones, line 133: Is there a reason not to do this in two lines? something like, 
response.getWriter().println(sendError(
"This query is not ....

src/main/webapp/query-control.js
```diff
@@ -80,20 +82,22 @@ function passQuery() {
...
-            // censusDataArray is a 2D array, where the first row is a
-            // header row and all subsequent rows are one piece of
-            // data (e.g. for a state or county)
-            const data = removeErroneousData(JSON.parse(response.data));
-            displayVisualization(data, description, location, isCountyQuery);
-            displayLinkToCensusTable(response.tableLink);
-            document.getElementById('more-info').innerText = '';
-          });
+      if (response.success) {
+        // censusDataArray is a 2D array, where the first row is a
```
shulijones, line 92: variable name is different now

src/main/java/QueryServlet.java
kyracq: Yeah, I had it in two lines but the linter asked for it to be like this, not really sure what the benefit is

PR #35, opened 2020-08-10
Add FAQ tooltip

shulijones:
Add a brief explanation of the site that appears when a user hovers over a ? button in the top right corner.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -29,9 +29,9 @@
              can mix and match between the lists to get answers to many 
              different questions. 
             <br><br>
-            The data you see comes from many different sources within the census.
+            The data you see comes from several sources within the census.
             Depending on your question, we choose the census data with the most 
-            information available. Most of our data comes from the annual 
+            information available. Most of our data come from the annual 
             American Community Survey; coming soon, if you ask about a year ending 
             in 0, you'll get information from the larger Decennial Survey instead.
           </p>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -24,7 +24,7 @@
       <h1>Census Viz
         <div class="tooltip">
           <i class="material-icons" id="faq">help_outline</i>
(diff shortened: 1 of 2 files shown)


src/main/webapp/index.html
```diff
@@ -20,7 +21,22 @@
 <body onload="registerServiceWorker(); createStateDropdownList();">
   <main>
     <div id="header">
-      <h1>Census Viz</h1>
+      <h1>Census Viz
+        <div class="tooltip">
+          <i class="material-icons" id="faq">help_outline</i>
```
nsri-google, line 26: Maybe add a one line summary of what censusviz is

src/main/webapp/index.html
```diff
@@ -20,7 +21,22 @@
...
+          <p class="tooltiptext">To use Census Viz, start typing in any box. The
```
nsri-google, line 27: maybe something like "select one of the available options from each drop down list and click the Green Go button. You can also start typing in any drop down to filter the options"

shulijones: That's a good suggestion, thank you. Done

src/main/webapp/index.html
shulijones: This is a good idea! I will discuss our "elevator pitch" with Kyra and Olivia and add in the best one-sentence summary we can come up with.

src/main/webapp/index.html
```diff
@@ -20,7 +21,22 @@
...
+            dropdown list will show you the options that are available. You
+            can mix and match between the lists to get answers to many
+            different questions. 
+            <br><br>
+            The data you see comes from several sources within the census.
+            Depending on your question, we choose the census data with the most
+            information available. Most of our data come from the annual
```
kyracq, line 34: most of our data comes

shulijones: Fixed, thanks

src/main/webapp/index.html
shulijones: Done

PR #36, opened 2020-08-10
Fix toggle and change color input styling

otobin:
The toggle between maps and amCharts should no longer appear when there is an error. The style of the color input should also be an oval without a border to better match the styling of our website. 


src/main/webapp/index.html
```diff
@@ -95,8 +95,10 @@ <h1>Census Viz</h1>
       <div id="map-title"></div>
       <div id="more-info"></div>
       <div id="colors">
-        <label for="set-color">Change the color of the visualization</label>
-        <input type="color" id="set-color" name="set-color" onchange="changeColor(this)">
+        <p>Change the color of the visualization</p>
+        <label id="set-color-label" for="set-color">
```
shulijones, line 99: Could you bring these two things closer together? Either having the color picker on the same line as the text, or having the two lines have a smaller gap between them. Right now the gap is kind of large (which makes it harder to figure out that you should be clicking on the color picker)

kyracq: ^^ I think it'd be nice to have the color pick on same line as text

otobin: Fixed!

PR #37, opened 2020-08-10
Add decennial data

shulijones:
Add data from the 2010 census, covering *all* counties, for queries about general population and population of men and women.

Data about people under/over-18 will require some more general code upgrades, which are the next item on my to-do list. Data other than population data, unfortunately, doesn't seem to be available from the decennial census. 

There appears to be a bug where sometimes a county is shown as having data in AmCharts, but not in GeoJson. I think that's a totally separate problem that this PR only highlights, but we can look into it (maybe tomorrow).

### DIFF
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -72,18 +72,18 @@ public class QueryServlet extends HttpServlet {
           ImmutableMap.of("all-ages", "S0201_119E,S0201_126E"));
 
   // Even more data available for population in 2010 when the decennial census happened
-  Map<String, Map<String, String>> queryToDataRowDecennial = 
+  Map<String, Map<String, String>> queryToDataRowDecennial =
       ImmutableMap.of(
           "live",
           ImmutableMap.of(
-            "all-ages",
-            "P001001",
-            "male",
-            "P012002",
-            "female",
-            "P012026")); /* TODO: can't do over/under-18 until after adding is implemented */
+              "all-ages",
+              "P001001",
+              "male",
+              "P012002",
+              "female",
+              "P012026")); /* TODO: can't do over/under-18 until after adding is implemented */
 
-  Map<String, String> tableNameToAbbrev = 
+  Map<String, String> tableNameToAbbrev =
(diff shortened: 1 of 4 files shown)


src/main/java/QueryServlet.java
```diff
@@ -71,34 +71,53 @@
...
+      // Decennial queries have overly specific tables, so always return a general one instead
+      return "https://data.census.gov/cedsci/table?tid=DECENNIALSF12010.P12";
+    }
     return "https://data.census.gov/cedsci/table?tid=ACS"
         + (dataRow.substring(0, 1).equals("K")
             ? "SE"
-            : (tableNameToAbbrev.get(dataTablePrefix.substring(1)) + "1Y"))
+            : (tableNameToAbbrev.get(
+                    dataTablePrefix.substring(dataTablePrefix.lastIndexOf("/") + 1))
+                + "1Y"))
```
kyracq, line 117: This could be indented to same level as prev line

src/main/java/QueryServlet.java
```diff
@@ -140,6 +159,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         && queryToDataRowPost2013.get(action).containsKey(personType)) {
       dataRow = queryToDataRowPost2013.get(action).get(personType);
     }
+    if (year == 2010
```
nsri-google, line 162: if this is guaranteed to be every 10 years you can do modulo 10.

src/main/java/QueryServlet.java
shulijones: You know, I thought so too, but the linter disagrees.

src/main/java/QueryServlet.java
shulijones: I was thinking about that, but from my preliminary poking around it seems like that rule will not necessarily hold fast - the only other modulo 10 years accessible by API are 1990 and 2000, and unfortunately it looks like 1990 has different mappings. So I left it this way for now, to be changed once I start working on getting 1990 and 2000 in.

PR #41, opened 2020-08-12
Reformat data in Java instead of JavaScript

shulijones:
Do data reformatting (percentage calculations etc.) in Java, and add support for additional kinds of reformatting (adding, subtracting, and chaining calculations). Client experience remains the same but data is more precise and there is more detailed data available for some queries.

I left it all in the same file right now for ease of use, but if it seems appropriate to break it out into a second Java file, I can do that as well. Edit: it is now in its own file.

### DIFF
--- a/src/main/java/DataFormatter.java
+++ b/src/main/java/DataFormatter.java
@@ -1,0 +1,125 @@
+package com.google.sps.data;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.io.InvalidObjectException;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.function.BiFunction;
+import java.util.List;
+import java.util.Map;
+
+// Reformats data from the QueryServlet according to a hard-coded list of instructions
+public class DataFormatter {
+
+  // Generic functions to combine numbers. Note that Math.round returns a long,
+  // but we can cast to int (max value ~2 billion) because our numbers
+  // will always be smaller than the U.S. population (~330 million)
+  private static BiFunction<Float, Float, Integer> add =
+      (Float a, Float b) -> (int) Math.round(a + b);
+  private static BiFunction<Float, Float, Integer> rightSubtract =
+      (Float a, Float b) -> (int) Math.round(a - b);
+  private static BiFunction<Float, Float, Integer> percent =
+      (Float a, Float b) -> (int) Math.round((a / 100.0) * b);
+
+  // Mapping data columns to the functions we want to execute on them --
+  // for example, if we are not able to access the population of children,
+  // instead we can do rightSubtract(total population, adult population)
+  private static Map<String, List<BiFunction>> dataRowToReformatFunction =
+      ImmutableMap.<String, List<BiFunction>>builder()
+          .put("DP05_0001E,DP05_0018E", ImmutableList.of(rightSubtract))
+          .put("S0201_119E,S0201_126E", ImmutableList.of(percent))
+          .put("K200104_001E,K200102_001E", ImmutableList.of(rightSubtract))
+          .put("S0701_C04_012E,S0701_C05_012E,S0701_C01_012E", ImmutableList.of(add, percent))
+          .put("S0701_C04_013E,S0701_C05_013E,S0701_C01_013E", ImmutableList.of(add, percent))
+          .put("S0201_125E,S0201_126E,S0201_119E", ImmutableList.of(add, percent))
+          .put("K200701_005E,K200701_006E", ImmutableList.of(add))
+          .put("K200701_004E,K200701_005E,K200701_006E", ImmutableList.of(add, add))
+          .put("S0201_124E,S0201_125E,S0201_126E,S0201_119E", ImmutableList.of(add, add, percent))
+          .put(
+              "S0701_C03_012E,S0701_C04_012E,S0701_C05_012E,S0701_C01_012E",
+              ImmutableList.of(add, add, percent))
+          .put(
+              "S0701_C03_013E,S0701_C04_013E,S0701_C05_013E,S0701_C01_013E",
+              ImmutableList.of(add, add, percent))
+          .put(
+              "P014001,P014021,P014022,P014042,P014043",
+              ImmutableList.of(rightSubtract, rightSubtract, rightSubtract, rightSubtract))
+          .build();
+
+  public static String reformatDataArray(String data, String dataIdentifier, boolean isCountyQuery)
+      throws InvalidObjectException {
+    if (!dataRowToReformatFunction.containsKey(dataIdentifier)) {
+      return data; // This data doesn't need reformatting
+    }
+    if ((dataRowToReformatFunction.get(dataIdentifier).size() + 1)
+        != dataIdentifier.split(",").length) {
+      // We should have enough functions listed to apply them to each group
+      // of two rows - (a, b) and then (their product, c), etc.
+      throw new InvalidObjectException(
+          "There aren't the right number of functions to apply to these data columns.");
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/QueryServlet.java
```diff
@@ -218,12 +244,128 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  // Math.round returns a long, but we can cast to int (max value ~2 billion) because our numbers
+  // will always be smaller than the U.S. population (~330 million)
+  BiFunction<Float, Float, Integer> add = (Float a, Float b) -> (int) Math.round(a + b);
+  BiFunction<Float, Float, Integer> rightSubtract = (Float a, Float b) -> (int) Math.round(a - b);
+  BiFunction<Float, Float, Integer> percent =
+      (Float a, Float b) -> (int) Math.round((a / 100.0) * b);
+
+  Map<String, List<BiFunction>> dataRowToReformatFunction =
+      ImmutableMap.<String, List<BiFunction>>builder()
+          .put("DP05_0001E,DP05_0018E", ImmutableList.of(rightSubtract))
```
otobin, line 280: Never worked with BiFunctions so I'm not entirely sure how these lines work. Are you mapping the two data tables to the appropriate function that will eventually be used to calculate the population value? Eg: On this line you map "DP05_0001E, DP05_0018E" to the function rightSubtract because we eventually want to subtract DP05_0018E from DP050001E?

shulijones: Yes, exactly! A Function (of which a BiFunction is a specific type) is just a way of passing a normal function/method as an object. Later in the code, I can call rightSubtract.apply(someNumber, someOtherNumber), and it will apply the function I wrote, (a, b) -> a - b, to subtract someOtherNumber from someNumber.
In this case, 0001E is the total population and 0018E is the adult population. So the subtraction will give us the child population.

src/main/java/QueryServlet.java
```diff
@@ -218,12 +244,128 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // Combine the original data, skipping the header row, by iterating over the functions
+    // given and successively combining the next datapoint with the previous result
+    List<BiFunction> numberCombiners = dataRowToReformatFunction.get(dataIdentifier);
+    for (int i = 0; i < numberCombiners.size(); i++) {
+      BiFunction numberCombiner = numberCombiners.get(i);
+      for (int j = 1; j < originalDataArray.size(); j++) {
+        List<String> originalDataRow = originalDataArray.get(j);
+        List<String> newDataRow = newDataArray.get(j);
+        if (newDataRow.get(1) == null || newDataRow.get(1).startsWith("-")) {
```
otobin, line 337: we use get(1) here because newDataRow only consists of two columns where the first column is the state name right? Like ["California", 120,000] and we would only want to access the 120,000

src/main/java/QueryServlet.java
```diff
@@ -218,12 +244,128 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+          // these get cleaned up in the JavaScript
+          newDataRow.set(1, "-1");
+        } else {
+          try {
+            int newValue =
+                (int)
+                    numberCombiner.apply(
+                        Float.parseFloat(newDataRow.get(1)), // Always replacing previous value
+                        Float.parseFloat(
+                            originalDataRow.get(i + 2))); // Moving along list of original values
```
otobin, line 348: Why i+2 here? If you're moving along the list of original values shouldn't it be j+1?

src/main/java/QueryServlet.java
```diff
@@ -218,12 +244,134 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+          // these get cleaned up in the JavaScript
+          newDataRow.set(1, "-1");
+        } else {
+          try {
+            int newValue =
+                (int)
+                    numberCombiner.apply(
+                        Float.parseFloat(newDataRow.get(1)), // Always replacing previous value
+                        Float.parseFloat(
+                            originalDataRow.get(i + 2))); // Moving along list of original values,
```
nsri-google, line 353: Maybe dataRowToReformatFunction should be Map<String, List<Pair<BiFunction,Integer>>> where the Integer value indicates which column from the original data to use for this operation. then you can use that value here instead of i+2

src/main/java/QueryServlet.java
```diff
@@ -218,12 +244,134 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        if (newDataRow.get(1) == null || newDataRow.get(1).startsWith("-")) {
+          // Sometimes the census API returns missing or negative numbers by mistake;
+          // these get cleaned up in the JavaScript
+          newDataRow.set(1, "-1");
+        } else {
+          try {
+            int newValue =
+                (int)
+                    numberCombiner.apply(
+                        Float.parseFloat(newDataRow.get(1)), // Always replacing previous value
```
nsri-google, line 351: It seems to be that you are performing string to float and then float to string conversions too many times. That is probably because of the way in the which the two loops are structured. If you switch them around then you work through rows of data and at the beginning you can convert to float once, apply all the transformations as you work through entries in numberCombiners and then finally write back as string once.

src/main/java/QueryServlet.java
shulijones: I'm a bit confused by what you mean, sorry. The value of i changes as we iterate through the list of BiFunctions - when we're at the first one, it's zero, then for the second one it's 1, etc. So as we go through all the functions, we also move across all the columns of the original data; we just skip the first two.

src/main/java/QueryServlet.java
shulijones: That's a good idea, thank you.

src/main/java/QueryServlet.java
nsri-google: I mean that you are using the position of the BiFunction in the list to determine the column it applies (with a +2). Instead you could explicitly specify that value along with the BiFunction, it would look something like 

put("S0201_124E,S0201_125E,S0201_126E,S0201_119E", ImmutableList.of(<add,3>, <add,4>, <percent,5>))

which translates to add data from column 3, then add data from column 4 and then use value in column 5 to calculate percentage. 

Currently since you always have them in order both these approaches work the same but image a case where you have to do the percent first and then the two adds then you would just rewrite this as 

put("S0201_124E,S0201_125E,S0201_126E,S0201_119E", ImmutableList.of(<percent,5>, <add,3>, <add,4>))

but your current approach has no simple way of supporting this. 

I don't want you to spend too much time on this so I am ok with keeping as is if you find this would be too much work.

shulijones: Ah, I understand now, thank you. I'm not opposed to implementing this but I do think it adds a good amount of complexity. I originally purposely rearranged all the lists of data so that they could proceed in order, as that seemed simpler to me at the time -- it makes no difference to the census API what order you request them in so in your example, you could just rearrange the list to get 3, 4, 5 in order again.

src/main/java/QueryServlet.java
shulijones: Done.

src/main/java/DataFormatter.java
```diff
@@ -0,0 +1,137 @@
...
+      finalDataRow.add(originalDataRow.get(numberCombiners.size() + 2)); // state ID
+      if (isCountyQuery) {
+        finalDataRow.add(originalDataRow.get(numberCombiners.size() + 3)); // county ID
+      }
+      finalDataArray.add(finalDataRow);
+    }
+
+    // Convert back into a JSON string for the JavaScript to read
+    String jsonData = gson.toJson(finalDataArray);
+    return jsonData;
```
anicr7, line 135: return gson.toJson(finalDataArray);

shulijones: Good point, thanks. Fixed

PR #42, opened 2020-08-13
Caching queries

kyracq:
Implement caching

Using [Objectify](https://github.com/objectify/objectify/wiki), a datastore API that also provides a global cache shared by all running instances of the application. By default all entities stay in the cache until it is full, not sure how entities are chosen to be evicted once cache is full.

Wasn't able to test speed improvements since this isn't deployed, but a local version can be ran with a datastore emulator that's started with
```console
gcloud beta emulators datastore start --host-port=localhost:8484
```

Also
- Switched to Mercator projection for state maps
- Relabeled reset zoom button

### DIFF
--- a/src/main/java/ObjectifyInit.java
+++ b/src/main/java/ObjectifyInit.java
@@ -1,0 +1,24 @@
+import com.google.cloud.datastore.DatastoreOptions;
+import com.googlecode.objectify.ObjectifyFactory;
+import com.googlecode.objectify.ObjectifyService;
+import com.google.sps.data.CensusData;
+import javax.servlet.ServletContextListener;
+import javax.servlet.ServletContextEvent;
+
+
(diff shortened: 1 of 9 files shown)


src/main/java/ObjectifyInit.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+
+public class ObjectifyInit implements ServletContextListener {
+  public void contextInitialized(ServletContextEvent event) {
+    ObjectifyService.init(
+        new ObjectifyFactory(
+            DatastoreOptions.newBuilder()
+                // This host is used for testing using a datastore emulator
+                .setHost("http://localhost:8484")
```
nsri-google, line 14: Will this cause a problem when you deploy to GCP?

PR #43, opened 2020-08-14
Allow user to enter non-state locations

shulijones:
Users can now type any location into the state dropdown (including cities, counties, store names, and zip codes - anything that could be successfully typed into Google Maps). When they do, they will see the visualization for the state containing their location, centered on the location itself.

I haven't yet thought of a good way to show the user that they can type anything they want into the location field (other than adding it to the FAQ); maybe we could style it a little differently? Would be interested in your thoughts here.

### DIFF
--- a/src/main/webapp/data-visualization.js
+++ b/src/main/webapp/data-visualization.js
@@ -1,9 +1,9 @@
 // Return geoJson for a given location. Load the geodata file if not already
 // loaded for the location.
-async function getGeoData(location, isCountyQuery) {
+async function getGeoData(locationInfo, isCountyQuery) {
   if (isCountyQuery) {
-    const abbrev = stateInfo[location].ISO.replace(/US-/, '').toLowerCase();
-    if (!stateInfo[location].geoJsonLoaded) {
+    const abbrev = stateInfo[locationInfo.number].ISO.replace(/US-/, '').toLowerCase();
+    if (!stateInfo[locationInfo.number].geoJsonLoaded) {
       await new Promise((resolve, reject) => {
         const script = document.createElement('script');
         document.body.appendChild(script);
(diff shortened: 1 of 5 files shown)


shulijones:
> Looks great! One suggestion is maybe it'd be good to display some info on the location being used? For example if you type in a city name that has matches in several different states it might be good to let the user know which one was found so they can make their query more specific if they were looking for info on somewhere else.
> 
> For letting the user know they can type any location one idea is having some example written below the dropdown?

@kyracq for your first suggestion -- currently you would know which state, since that's the one that would load afterwards. Do you mean more like, in the moment/responsively?

Your second suggestion is good, thank you!! Are you imagining it would be like, underneath the line but separate from the dropdown, or an additional element within the dropdown list?

shulijones:
> I got some weird results when I tried typing in city names like Baltimore, Las Vegas, and Indianapolis. For both the amcharts and maps, the center of the map changed but all of the data was still overlayed on California. Also, there are a couple of errors when I change the color of the map.
> 
> As for indicating to the user that they can type in any location, you can implement an autocomplete box as part of the places API (https://developers.google.com/places/web-service/autocomplete). This would be really helpful for city names that exist in different states and also it would indicate to the user that they can type in anything.

That's very interesting, I don't get those results! For me Baltimore shows Maryland and Las Vegas shows Nevada. I wonder if it's because you're in California...? Could you try doing something like "Las Vegas, Nevada" and seeing if that fixes it?

I also get the color changing errors, so thank you for pointing that out. I didn't think to test that and I'll see about pushing fixes to that as soon as I can.

The autocomplete box is a great idea! It looks like that might be a bit more complicated (since you have to parse through the information the API returns), but I'll get started working on it.

PR #44, opened 2020-08-14
URL addressibility, logging in

kyracq:

Note: Some lines in the ObjectifyInit need to be commented/uncommented when switching between deploying and running
- Next step: save queries when logged in


src/main/java/LoginServlet.java
```diff
@@ -0,0 +1,35 @@
...
+
+    if (!userService.isUserLoggedIn()) {
+      jsonResponse.addProperty("loggedIn", false);
+      // After login redirect user to nickname-input page
+      String loginUrl = userService.createLoginURL("/");
+      jsonResponse.addProperty("loginUrl", loginUrl);
+    } else {
+      jsonResponse.addProperty("loggedIn", true);
+      String logoutUrl = userService.createLogoutURL("/");
+      jsonResponse.addProperty("logoutUrl", logoutUrl);
```
shulijones, line 26: I can't seem to log out. When I try, the page says "Redirecting you to https://accounts.google.com/Logout?continue=https://uc.appengine.google.com/_ah/logout%3Fcontinue%3Dhtt..." (there's much more to this URL), then it opens my Chrombook's settings. But the settings don't have a way to log out as far as I can tell.

Also, it would be nice if it were clear "how" people were logging in when they click log in - when I clicked, the page just refreshed and said "log out". There was no indication that it was my google account that had been logged in.

src/main/java/LoginServlet.java
```diff
@@ -0,0 +1,35 @@
...
+public class LoginServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JsonObject jsonResponse = new JsonObject();
+    UserService userService = UserServiceFactory.getUserService();
+
+    if (!userService.isUserLoggedIn()) {
+      jsonResponse.addProperty("loggedIn", false);
+      // After login redirect user to nickname-input page
+      String loginUrl = userService.createLoginURL("/");
```
shulijones, line 21: I didn't get redirected to a nickname input page after I logged in, the homepage just refreshed.

src/main/webapp/index.html
```diff
@@ -15,15 +15,18 @@
...
 </head>
 
-<body onload="registerServiceWorker(); createStateDropdownList(); loadAppropriateIcon(false);">
+<body
+  onload="registerServiceWorker(); createStateDropdownList(); loadAppropriateIcon(false); getAndDisplayLoginStatus();">
   <main>
     <div id="header">
-      <h1><a href="/">Census Viz</a>
+      <h1><a href="/">Census Viz</a></h1>
+      <div id="header-buttons">
```
shulijones, line 29: The buttons used to all be vertically aligned with each other, and now they're not. I'm not sure exactly why but it definitely is off.

src/main/webapp/login-status.js
```diff
@@ -0,0 +1,25 @@
...
+// Set HTML content according to whether user is logged in.
+function displayLoginStatus(loginStatus) {
+  const loginStatusElement = document.getElementById('login-status');
+
+  if (loginStatus.loggedIn) {
+    loginStatusElement.innerHTML =
+    `<a href=${loginStatus.logoutUrl}>Log out</a>`;
+  } else {
+    const loginLink = `<a href=${loginStatus.loginUrl}>Log in</a>`;
+    loginStatusElement.innerHTML = loginLink;
```
shulijones, line 17: Is there a reason not to do this in one line, like the logout one does?

src/main/webapp/index.html
shulijones: Imo there's also a bit too much space above the header (more than is below the words/icons), so that it doesn't look as symmetrical, but I think that's more subjective

src/main/java/LoginServlet.java
otobin: ^^

src/main/java/LoginServlet.java
otobin: ^^

src/main/java/LoginServlet.java
kyracq: This comment was left over since I used basically the same code from my portfolio, there isn't a nickname input page, removed comment

src/main/webapp/login-status.js
kyracq: Nope, fixed

PR #45, opened 2020-08-18
Login

kyracq:
- Add URL admissibility (queries in URL)
- Logging in
- Click to query state on all states amcharts map


src/main/webapp/index.html
```diff
@@ -56,39 +55,41 @@ <h1><a href="/">Census Viz</a>
         <div class="tooltip">
           <i class="material-icons" id="location-off-icon" onclick="loadAppropriateIcon(true)">location_off</i>
```
shulijones, line 56: The buttons look great to me, thanks for fiddling with them!

src/main/webapp/index.html
```diff
@@ -148,8 +147,10 @@ <h1><a href="/">Census Viz</a>
       <div id="data-table"></div>
     </div>
     <footer>
-      Made by Kyra Acquah, Shuli Jones, and Olivia Tobin.
+      <div>Made by Kyra Acquah, Shuli Jones, and Olivia Tobin.</div>
+      <div id="sign-out-btn"><a href="/" onclick="signOut();">Sign out</a></div>
```
shulijones, line 151: Would it be possible to give this button the same styling as the sign in button? If not, no worries

src/main/webapp/login-status.js
```diff
@@ -0,0 +1,14 @@
...
+  console.log('Name: ' + profile.getName());
+  console.log('Image URL: ' + profile.getImageUrl());
+  console.log('Email: ' + profile.getEmail());
+}
+
+// Sign user out
+function signOut() {
+  const auth2 = gapi.auth2.getAuthInstance();
+  auth2.signOut();
+}
```
shulijones, line 14: After I sign out, I get redirected to the home page. Would it be possible to get redirected back to the page I was already on?

src/main/webapp/login-status.js
```diff
@@ -0,0 +1,14 @@
+// Get signed in user info
+function onSignIn(googleUser) {
+  const profile = googleUser.getBasicProfile();
+  const idToken = googleUser.getAuthResponse().id_token;
+  console.log('Name: ' + profile.getName());
+  console.log('Image URL: ' + profile.getImageUrl());
+  console.log('Email: ' + profile.getEmail());
```
shulijones, line 7: Probably want to delete these

src/main/webapp/query-control.js
```diff
@@ -35,45 +35,67 @@ function clearPreviousResult() {
...
-async function passQuery() {
-  clearPreviousResult();
+function getFetchUrl(personType, action, location, year) {
+  return '/query?person-type=' + personType +
+    '&action=' + action +
+    '&location=' + location +
+    '&year=' + year;
+}
+
+
```
shulijones, line 45: Delete one of the two empty lines

src/main/webapp/data-visualization.js
```diff
@@ -72,8 +72,10 @@ function displayAmChartsMap(data, locationInfo, description, geoData, color) {
     chart.goHome();
   });
 
+  const isCountyQuery = data[0].id.indexOf('US') === -1;
+
   // Albers projection for state level, Mercator for county level
-  if (data[0].id.indexOf('US') !== -1) {
+  if (!isCountyQuery) {
```
shulijones, line 78: Nit: any reason not to swap these and avoid negating the boolean?

shulijones (approved):
This looks great, Kyra, thank you! Code LGTM, all my suggestions are minor.

src/main/webapp/index.html
otobin, line 151: nit: maybe hide the sign out button when the user is already signed out?

kyracq: I tried to do this this morning but it didn't seem like I was able to modify the style/display of the sign in button.

src/main/webapp/data-visualization.js
kyracq: Fixed!

src/main/webapp/query-control.js
kyracq: Done.

src/main/webapp/login-status.js
kyracq: Done

src/main/webapp/login-status.js
kyracq: Fixed!

src/main/webapp/index.html
kyracq: Done, currently looks the same except without the Google logo, open to thoughts on whether I should also add the logo

shulijones: Looks great, thanks! I think either way's fine, nbd

src/main/webapp/login-status.js
shulijones: Thanks! Nit: would it be possible to also make the page refresh? When I click the sign-out button it kind of feels like nothing happens, and I have to scroll back up and notice that the sign-in button now says "sign in" instead of "signed in". This is really a nit though

src/main/java/ObjectifyInit.java
```diff
@@ -16,7 +16,8 @@ public void contextInitialized(ServletContextEvent event) {
                 .build()
                 .getService()));
     // To deploy, everything above this must be commented out
-    // and replaced with only ObjectifyService.init();
+    // and line directly below show be uncommented
+    // ObjectifyService.init();
```
nsri-google, line 20: I think this is not a good situation to be in where you have to comment out lines of code to make it work locally versus when deployed. We should talk about a cleaner way to do this.

nsri-google (approved):

PR #47, opened 2020-08-19
Autocomplete location

shulijones:
As the user types in the location box, offer autocompleted suggestions in addition to the hardcoded state options.

I've spent a good bit of time trying to get the autocomplete to merge nicely with the state dropdown, with mostly good results. One remaining issue is that the dropdown box does a little shake every time the autocomplete resets, which can be annoying. Also, I've got a delay on the autocomplete so that it doesn't try over and over as you type, and instead runs at intervals; getting the interval just right is tricky, so let me know if you think it's too long or too short.

### DIFF
--- a/src/main/webapp/query-control.js
+++ b/src/main/webapp/query-control.js
@@ -241,8 +241,9 @@ function setupAutocompleteLocation() {
       return;
     }
 
-    predictions.splice(5); // just show top 5 predictions
-    // each time we display suggestions, show only the top 5, and then the state dropdown
+    // Each time we display suggestions, show only the top 5, and then the
+    // normal state dropdown afterwards
+    predictions.splice(5);
     let resultsHtml = [];
     predictions.forEach(function(prediction) {
       resultsHtml.push(
--- a/src/main/webapp/query-control.js
+++ b/src/main/webapp/query-control.js
@@ -119,6 +119,7 @@ async function passQuery() {
     });
 }
 
+
 // Remove incorrect data returned by the census API
 // such as Puerto Rico data and negative numbers
 function removeErroneousData(dataArray) {
@@ -198,6 +199,11 @@ function getSortedStateInfoArray() {
   return stateInfoArray;
 }
 
+function setupLocationDropdown() {
+  createStateDropdownList();
+  createAutocompleteLocation();
+}
+
 // Append all locations to the location dropdown element.
 async function createStateDropdownList() {
   const defaultLocation = await getUserState();
@@ -213,6 +219,28 @@ async function createStateDropdownList() {
     optionElem.value = value.name;
     optionElem.setAttribute('data-value', value.number);
     datalist.appendChild(optionElem);
+  });
+}
+
+// Set up an autocomplete dropdown to suggest places to users
+function createAutocompleteLocation() {
+  const input = document.getElementById('location-list');
+  const autocomplete = new google.maps.places.Autocomplete(input);
+  autocomplete.setFields(
+    ['address_components', 'geometry', 'icon', 'name']);
+
+  autocomplete.addListener('place_changed', function() {
+    //infowindow.close();
+    //marker.setVisible(false);
+    const place = autocomplete.getPlace();
+    if (!place.geometry) {
+      // User entered the name of a Place that was not suggested and
+      // pressed the Enter key, or the Place Details request failed.
+      window.alert("No details available for input: '" + place.name + "'");
+      return;
+    }
+    const input = document.getElementById('location-list');
+    input.value = place.name;
   });
 }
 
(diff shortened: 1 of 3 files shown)


src/main/webapp/query-control.js
```diff
@@ -119,6 +121,7 @@ async function passQuery() {
     });
 }
 
+
```
kyracq, line 124: Can delete this extra line

src/main/webapp/query-control.js
```diff
@@ -216,6 +224,75 @@ async function createStateDropdownList() {
...
+// Set up an autocomplete dropdown, attached to the main location input
+// and states dropdown, to suggest places to users as they type
+function setupAutocompleteLocation() {
+  const autocompleteInput = document.getElementById('location-list');
+  const autocompleteResults = document.getElementById('location');
+  const service = new google.maps.places.AutocompleteService();
+  const defaultLocationOptions = autocompleteResults.innerHTML; // all states
+
+  function displaySuggestions(predictions, status) {
+    if (status != google.maps.places.PlacesServiceStatus.OK) {
```
kyracq, line 236: Not sure how the status value works but could use a !== if possible

src/main/webapp/query-control.js
```diff
@@ -216,6 +224,75 @@ async function createStateDropdownList() {
...
+      alert(
```
kyracq, line 237: When I start with the default location, click the dropdown, and click the default each U.S. state again, I get the zero results error. (Or if I just try typing "each "). Is there a way to not have the alert pop up for zero results and just show the dropdown with only states? Also, the input clears when there are no autocomplete results, though the input would be a business if it doesn't show in the suggestions, I'm wondering if users should be allowed to input anything?

src/main/webapp/query-control.js
```diff
@@ -216,6 +224,75 @@ async function createStateDropdownList() {
...
+  // on the predictions it generated
+  autocompleteInput.addEventListener('input', debounce(function() {
+    const value = this.value;
+    value.replace('"', '\\"').replace(/^\s+|\s+$/g, ''); // Trim whitespace
+    service.getPlacePredictions({
+      'input': value,
+      'types': ['(regions)'], // no businesses, just cities counties etc.
+      'componentRestrictions': {'country': 'us'}, // USA only
+    }, displaySuggestions);
+  }, 250)); // wait 250 ms before sending the request so they don't overload
```
kyracq, line 271: I think a slightly shorter delay may be better, so users can see available suggestions before continuing typing. I also tried this without the delay locally and it seems to work fine with no delay, what issues did you run in to without the delay? I do notice that the shake becomes more obvious.

src/main/webapp/query-control.js
shulijones: Done, thanks

src/main/webapp/query-control.js
shulijones: Oh yeah probably, thanks

src/main/webapp/query-control.js
shulijones: Yeah, I added the delay to try to get rid of the shake. Maybe I'm a bit sensitive but for me it makes it pretty hard to read the results list when it's at that level of shaking

shulijones: Tried out some options - 175 or 200 seem like they might be the sweet spot here for shaking vs too unresponsive.

src/main/webapp/query-control.js
shulijones: Ah, that's a good point. I can just delete this if-clause and leave the next one that returns when there aren't any predictions. 
I'm not sure what you mean by your second question, sorry, could you elaborate?

kyracq: Oh after checking again the input comes back after you click away, as in if I type a non-location in, the zero results alert pops up, then I click ok, then the location input is blank, but then I click somewhere else and the non-location I typed comes back. Was wondering if the input can come back right after clicking OK

shulijones: Ah, gotcha. I think I deleted the alert entirely, so this shouldn't be a problem any more. I'll deploy the latest version now so you can take a look.

kyracq: Looks and works great now, thanks!

src/main/webapp/query-control.js
```diff
@@ -216,6 +223,68 @@ async function createStateDropdownList() {
...
+  // on the predictions it generated
+  autocompleteInput.addEventListener('input', debounce(function() {
+    const value = this.value;
+    value.replace('"', '\\"').replace(/^\s+|\s+$/g, ''); // Trim whitespace
+    service.getPlacePredictions({
+      'input': value,
+      'types': ['(regions)'], // no businesses, just cities counties etc.
+      'componentRestrictions': {'country': 'us'}, // USA only
+    }, displaySuggestions);
+  }, 175)); // wait this many ms to send request so they don't overload
```
otobin, line 263: I didn't see earlier versions but I think 175 is good, I'm not seeing any shaking and I can both type fast and see results if I wait.

PR #50, opened 2020-08-20
Show other visualizations

otobin:
If a user is currently logged in, they should be able to see a list of links of all the visualizations they've viewed. History should save across login/logout sessions and should appear on the homepage even if a query hasn't been made. 


src/main/java/CensusData.java
```diff
@@ -26,4 +27,5 @@ public String getData() {
   public String getTableLink() {
     return tableLink;
   }
+
```
shulijones, line 30: Delete?

src/main/java/CensusData.java
```diff
@@ -3,6 +3,7 @@
 import com.googlecode.objectify.annotation.Cache;
 import com.googlecode.objectify.annotation.Entity;
 import com.googlecode.objectify.annotation.Id;
+import java.util.List;
```
shulijones, line 6: Doesn't look like you're using this

src/main/java/HistoryServlet.java
```diff
@@ -0,0 +1,68 @@
...
+    historyEntity.setProperty("personType", personType);
+    historyEntity.setProperty("action", action);
+    historyEntity.setProperty("location", location);
+    historyEntity.setProperty("year", year);
+    historyEntity.setProperty("userId", userId);
+    
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(historyEntity);
+
+    HistoryElement requestHistoryElement = new HistoryElement(userId,
```
shulijones, line 41: At this point you switch to responding with the whole list of stored queries, right? Maybe add a comment clarifying, to go with your original one saying you're putting the current query in datastore

src/main/java/QueryServlet.java
```diff
@@ -145,6 +147,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.setContentType("application/json;");
 
     String query = personType + action + location + yearStr;
+    ArrayList<String> queryList = new ArrayList<String>(
```
shulijones, line 150: Doesn't seem like you're using this

src/main/webapp/query-control.js
```diff
@@ -35,13 +35,82 @@ function clearPreviousResult() {
...
+
+  // change back the action to be gramatically correct in the title
+  const gramaticallyCorrectAction = new Map();
+  gramaticallyCorrectAction.set(
+        'live', 'lived in',
+      ).set(
+        'work', 'worked in',
+      ).set(
+        'moved', 'moved to',
+      );
```
shulijones, line 76: Could you use the getTitle function to do this instead?

src/main/webapp/query-control.js
```diff
@@ -35,13 +35,82 @@ function clearPreviousResult() {
...
-function getFetchUrl(personType, action, location, year) {
-  return '/query?person-type=' + personType +
+function getFetchUrl(servlet, personType, action, location, year) {
+  return '/' + servlet + '?person-type=' + personType +
     '&action=' + action +
     '&location=' + location +
     '&year=' + year;
 }
 
+function getTitle(personType, location, year, locationInput, actionInput) {
```
shulijones, line 45: This is a nice modularization. There's a spot later in the code that could use it as well (currently the function is basically just copied in two places).

src/main/webapp/query-control.js
```diff
@@ -35,13 +35,82 @@ function clearPreviousResult() {
...
+  gramaticallyCorrectAction.set(
+        'live', 'lived in',
+      ).set(
+        'work', 'worked in',
+      ).set(
+        'moved', 'moved to',
+      );
+  fetch(fetchUrl).then((response) => response.json()).then(
+    function (jsonResponse) {
+      jsonResponse.forEach((historyElement) => {
```
shulijones, line 79: A little more commenting here would be helpful

src/main/webapp/index.html
```diff
@@ -167,6 +167,7 @@ <h1><a href="/">Census Viz</a></h1>
       <br>
       <div id="data-table"></div>
     </div>
+    <div id="history"></div>
```
shulijones, line 170: This is irrelevant to this line, but I wanted to comment in index.html. Can you add getHistory to the onload? Right now you have to make a query before you can see your previous history (but it maintains your history from other tabs you have open, which is great!) I'd like to see my history even before I make any queries, if I'm logged in.

Note that this might require you to add a check to the very beginning of getHistory that returns without doing anything if the user is logged out. But I think that would be a smart thing to have anyway.

src/main/java/HistoryElement.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public boolean equals(Object other) {
+    HistoryElement otherElement = (HistoryElement)other;
+    String otherPersonType = otherElement.getPersonType();
+    String otherAction = otherElement.getAction();
+    String otherLocation = otherElement.getLocation();
+    String otherYear = otherElement.getYear();
+    return ((this.personType.equals(otherPersonType)) && 
+        (this.action.equals(action)) &&
```
kyracq, line 46: I think you're accidentally comparing to action instead of otherAction

src/main/java/HistoryServlet.java
```diff
@@ -0,0 +1,68 @@
...
+    historyEntity.setProperty("year", year);
+    historyEntity.setProperty("userId", userId);
+    
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(historyEntity);
+
+    HistoryElement requestHistoryElement = new HistoryElement(userId, 
+      personType, action, location, year);
+
+    Query query = new Query("historyEntity");
```
kyracq, line 44: You could try adding a [filter](https://cloud.google.com/appengine/docs/standard/java/datastore/queries#filters) to this query to only get the history for a particular userId so you don't have to check the userId in the for loop later

src/main/webapp/login-status.js
```diff
@@ -49,3 +49,18 @@ function loginInit() {
...
+  const auth2 = gapi.auth2.getAuthInstance();
+  return auth2.isSignedIn.get();
+}
+
+
+function getUserId() {
+  const auth2 = gapi.auth2.getAuthInstance();
+  if (isUserSignedIn()) {
+    var profile = auth2.currentUser.get().getBasicProfile();
+    return profile.getId();
```
kyracq, line 64: Google [recommends against](https://developers.google.com/identity/sign-in/web/backend-auth) using this user id but the id_token I get in onSignIn instead.

src/main/webapp/index.html
kyracq: ^^ you might be able to do this by getting history in loginInit(). Also, if you're signed out then sign in on a query result history doesn't show until you make a new query, it might be good to also get history in the onSignIn function

src/main/webapp/query-control.js
```diff
@@ -35,13 +35,82 @@ function clearPreviousResult() {
...
+        if (historyElement !== null) {
+          let historyTextNode;
+          if (historyElement.location !== 'state') {
```
kyracq, line 82: You could shorten this/remove some repetition by doing something like
location = historyElement.location === 'state' ? 'Each U.S. state' : stateInfo[historyElement.location].name
and then call getTitle with that location variable

src/main/webapp/query-control.js
```diff
@@ -35,13 +35,82 @@ function clearPreviousResult() {
...
+          const breakElement = document.createElement('br');
+          historyContainer.appendChild(breakElement);
+        }
+      })});
+}
+
+// Given a history element, return the appropriate url
+function getHistoryUrl(historyElement) {
+  const host = window.location.origin;
+  const hash = '/#person-type=' + historyElement.personType +
```
kyracq, line 107: You could use fetchUrl here by either modifying fetchUrl or calling fetchUrl as is and then removing the servlet part

src/main/java/CensusData.java
otobin: yeah probably a good idea

src/main/java/CensusData.java
otobin: Good point. I was earlier but later deleted that functionality.

src/main/java/QueryServlet.java
otobin: true

src/main/java/HistoryServlet.java
otobin: good point! done

src/main/webapp/query-control.js
otobin: moved gramaticallycorrectaction to getTitle

src/main/webapp/query-control.js
otobin: Good point! Fixed

src/main/java/HistoryElement.java
otobin: true! nice catch

src/main/webapp/query-control.js
otobin: Good point! I did the latter

src/main/java/HistoryServlet.java
otobin: I tried doing this earlier but I couldn't get it to compile because I wasn't constructing the filter properly. The code in the documentation works!

src/main/webapp/query-control.js
otobin: get HistoryUrl now calls to getFetchUrl to eliminate repeated code

src/main/webapp/query-control.js
otobin: good idea! done

src/main/webapp/login-status.js
```diff
@@ -15,8 +21,13 @@ function signOut() {
 
 // Get whether user is signed in
 function getLoginStatus() {
-  const auth2 = gapi.auth2.getAuthInstance();
-  return auth2.isSignedIn.get();
+  const auth2 = gapi.auth2;
```
kyracq, line 24: Noticed while working on history UI this should be a let since it's reassigned in else block

src/main/java/HistoryServlet.java
```diff
@@ -0,0 +1,73 @@
...
+    String location = request.getParameter("location");
+    String year = request.getParameter("year");
+    String userId = request.getParameter("user-id");
+
+    Entity historyEntity = new Entity("historyEntity");
+    historyEntity.setProperty("personType", personType);
+    historyEntity.setProperty("action", action);
+    historyEntity.setProperty("location", location);
+    historyEntity.setProperty("year", year);
+    historyEntity.setProperty("userId", userId);
```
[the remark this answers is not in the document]
anicr7: Agree here, some of these temp vars are probably not needed. You can directly use it in the SetProperty calls.

src/main/webapp/login-status.js
otobin: nice catch! fixed.

src/main/java/HistoryServlet.java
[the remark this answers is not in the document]
otobin: done!

src/main/java/QueryServlet.java
```diff
@@ -11,6 +11,8 @@
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.Map;
+import java.util.ArrayList;
+import java.util.Arrays;
```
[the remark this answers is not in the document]
otobin: nope! deleted

src/main/java/HistoryElement.java
```diff
@@ -0,0 +1,50 @@
+package com.google.sps.data;
+
+public class HistoryElement {
+  String id;
```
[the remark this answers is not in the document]
otobin: true! done.

PR #52, opened 2020-08-20
Don't require commenting/uncommenting

kyracq:
We'll no longer have to comment/uncomment to deploy. We just need to have three env variables set:
```console
export DATASTORE_PROJECT_ID=censusviz
export DATASTORE_USE_PROJECT_ID_AS_APP_ID=true
export DATASTORE_EMULATOR_HOST=localhost:8484
```
We can add these to ~/.bashrc so they're always set

### DIFF
--- a/src/main/java/ObjectifyInit.java
+++ b/src/main/java/ObjectifyInit.java
@@ -7,17 +7,7 @@ import javax.servlet.ServletContextListener;
 
 public class ObjectifyInit implements ServletContextListener {
   public void contextInitialized(ServletContextEvent event) {
-    ObjectifyService.init(
-        new ObjectifyFactory(
-            DatastoreOptions.newBuilder()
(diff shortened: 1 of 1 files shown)


nsri-google:
> We'll no longer have to comment/uncomment to deploy. We just need to have three env variables set:
> 
> ```
> export DATASTORE_PROJECT_ID=censusviz
> export DATASTORE_USE_PROJECT_ID_AS_APP_ID=true
> export DATASTORE_EMULATOR_HOST=localhost:8484
> ```
> 
> We can add these to ~/.bashrc so they're always set

Maybe you can create a script that sets these and then starts the datastore emulator

nsri-google:
Actually it looks like there is already a way to do that - https://cloud.google.com/datastore/docs/tools/datastore-emulator#automatically_setting_the_variables

PR #54, opened 2020-08-24
Year slider

shulijones:
Add a year slider to change the year being visualized.

This is a very simple implementation; it just sends a new query every time the user changes the year on the slider.
(Note: we are currently discussing the UI, including the positioning of all the elements above the map and the coloring, so this may change later.)

### DIFF
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -118,6 +118,35 @@ input[type=submit]:hover {
   transition: all .4s ease-in-out;
 }
 
+#year-slider {
+  width: 100%;
+}
+
+#set-year {
+  -moz-appearance: none;
+  -webkit-appearance: none;  /* Override default style for Chrome and Mozilla */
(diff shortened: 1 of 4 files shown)


src/main/webapp/style.css
```diff
@@ -118,6 +118,34 @@ input[type=submit]:hover {
...
+  border-radius: 25px;
+  height: 10px;
+  outline: none;
+  width: 40%;
+}
+
+/* The handle for the set-year slider (once for Chrome, once for Mozilla) */
+#set-year::-webkit-slider-thumb {
+  -webkit-appearance: none;
+  background: #0dbd00;
```
otobin, line 138: not sure if this is just me, but when I changed the color of the map the green button sometimes clashed with the different color scheme. I was thinking it might be cool to change the slider to match the color picker, but I'm curious to hear other thoughts.

PR #55, opened 2020-08-24
History ui

kyracq:

Notes:
- Query History doesn't show up immediately after sign in, because doing this led to getHistory being called twice and duplicates in the carousel
- When on first page of slider/carousel, the back button should take you to the last page (when on last page forward button should take you to first page)
- We don't have timestamps in history elements yet, so for now I just reverse the history array which seems to give approximately the desired result


src/main/webapp/style.css
```diff
@@ -307,3 +307,46 @@ input[type="color"] {
   padding: 8px;
   height: 18px;
 }
+
```
shulijones, line 310: We may want to start dividing this document up into sections using big comments as headers; we've got a lot of different groups of css going on here now.

kyracq: Done!

src/main/webapp/login-status.js
```diff
@@ -53,13 +41,13 @@ function loginInit() {
...
-      gapi.signin2.render('sign-in-btn', {
-        'onsuccess': onSignIn,
-      });
+      gapi.signin2.render('sign-in-btn');
       displayLoginStatus();
+      toggleHistory(false);
     }).then(() => {
-      if (getLoginStatus()) {
-        getHistory(getUserId());
+      const isLoggedIn = getLoginStatus();
```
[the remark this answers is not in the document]
kyracq: Fixed

src/main/webapp/query-control.js
```diff
@@ -289,6 +330,13 @@ function replaceValueIfEmpty(dataListId) {
   }
 }
 
+// Change the year of data being visualized
+function changeYear(yearParam) {
+  const year = yearParam.value;
+  document.getElementById('year-list').value = year;
```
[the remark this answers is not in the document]
kyracq: @shulijones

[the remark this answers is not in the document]
shulijones: I can change that! Thanks.

src/main/webapp/query-control.js
```diff
@@ -76,47 +76,86 @@ function putHistory(personType, action, location, year, userId) {
...
+function createSlide(title, url, location) {
+  const historyItem = document.createElement('li');
+  historyItem.className = 'splide__slide';
+  const link = document.createElement('a');
+  link.href = url;
+  const titleNode = document.createTextNode(title);
+  const imgWrapper = document.createElement('div');
+  imgWrapper.className = 'img-wrapper';
+  const image = document.createElement('img');
+  image.alt = 'Small U.S. map';
```
shulijones, line 88: now that the map is different depending on location, you could append ' of ' + stateInfo[location].name for county queries

src/main/webapp/style.css
```diff
@@ -118,34 +132,40 @@ input[type=submit]:hover {
...
+#query-form {
+  display: inline-flex;
 }
 
+#query {
+  text-align: center;
+}
+
+/* Visualzations */
```
shulijones, line 168: nit: typo here

kyracq:
> LGTM! I love the static maps for each state. One nitpick is that when you sign out and then sign in again without the page reloading, the history only shows up after you've made a query. Because history disappears instantaneously it might be good to also have it appear again instantaneously.

Yeah, I also agree that'd be ideal but for some reason doing that caused duplicates in the carousel

PR #56, opened 2020-08-24
Add button to edit title

shulijones:
Allow the user to edit the title of the amcharts map.

The title they choose will be saved when they download the image, but not on page reload. The user can "get back" our default title by reloading the page, asking the query again, or simply entering a blank title into the box.

Notes:
-I mostly didn't add a lot of formatting to these buttons because we are discussing doing a sitewide redesign of our buttons right now anyway.
-This PR also includes an update to the year dropdown to label 2010 as a special year, and an update to the amcharts export button so it says Download explicitly. Both are very small changes.

### DIFF
--- a/src/main/webapp/data-visualization.js
+++ b/src/main/webapp/data-visualization.js
@@ -91,26 +91,26 @@ function displayAmChartsMap(
         {
           'label': 'Image',
           'menu': [
-            {'type': 'png', 'label': 'PNG' },
-            {'type': 'jpg', 'label': 'JPG' },
-            {'type': 'svg', 'label': 'SVG' },
-            {'type': 'pdf', 'label': 'PDF' },
+            {'type': 'png', 'label': 'PNG'},
+            {'type': 'jpg', 'label': 'JPG'},
+            {'type': 'svg', 'label': 'SVG'},
+            {'type': 'pdf', 'label': 'PDF'},
           ],
         }, {
           'label': 'Data',
           'menu': [
-            {'type': 'json', 'label': 'JSON' },
-            {'type': 'csv', 'label': 'CSV' },
-            {'type': 'xlsx', 'label': 'XLSX' },
-            {'type': 'html', 'label': 'HTML' },
-            {'type': 'pdfdata', 'label': 'PDF' },
+            {'type': 'json', 'label': 'JSON'},
+            {'type': 'csv', 'label': 'CSV'},
+            {'type': 'xlsx', 'label': 'XLSX'},
+            {'type': 'html', 'label': 'HTML'},
+            {'type': 'pdfdata', 'label': 'PDF'},
           ],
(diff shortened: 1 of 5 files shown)


shulijones:
> LGTM! I think it would be cool if when the user navigated back to the page that they changed the title for, it was retained. It would also be cool if the title was the name of the downloaded image. These two are def not high priority though

The second one is very easy, the first one is very hard. So I did the second one only for now :)

src/main/webapp/query-control.js
```diff
@@ -289,6 +295,16 @@ function replaceValueIfEmpty(dataListId) {
   }
 }
 
+function showEditTitle() {
```
kyracq, line 298: Maybe 'edit title' button could be hidden when looking at geoJson, since that doesn't have a title. Though it may also be useful to edit tile there and see it when you switch back to amcharts

shulijones: Ah, that's a good point! I think it would be more confusing to users to see the button when there isn't a title, so I'll hide it.

src/main/webapp/data-visualization.js
```diff
@@ -85,7 +86,34 @@ function displayAmChartsMap(
...
+        {
+          'label': 'Image',
+          'menu': [
+            {'type': 'png', 'label': 'PNG'},
+            {'type': 'jpg', 'label': 'JPG'},
+            {'type': 'svg', 'label': 'SVG'},
+            {'type': 'pdf', 'label': 'PDF'},
+          ],
+        }, {
+          'label': 'Data',
```
nsri-google, line 103: I do not see this in the UI

src/main/webapp/query-control.js
```diff
@@ -290,6 +296,18 @@ function replaceValueIfEmpty(dataListId) {
   }
 }
 
+// Show the div that allows for AmCharts map title editing
+function showEditTitle() {
+  document.getElementById('edit-title').style.display = 'inline';
```
nsri-google, line 301: you should pre-populate the textbox with the current title

src/main/webapp/data-visualization.js
shulijones: Oh right. This is the default amcharts menu options, but for maps, they actually don't show the data because it's set up as a separate polygon series, not as the chart data. I'll tag @kyracq and ask her if she knows if there's a way to set our data as chart.data and not just polygonSeries.data, but otherwise, I can remove this set of options entirely.

src/main/webapp/query-control.js
shulijones: Good idea! Done.

src/main/webapp/data-visualization.js
shulijones: Kyra found a way :) Should be done now.

PR #57, opened 2020-08-25
Show recommendations

otobin:
User should now be able to see their history and recommended queries based on their search history. Not currently deployed, but should run fine locally.

The recommendation algorithm works by creating a hash map for each field in the query that maps the option to the frequency. Then, it sorts each map in descending order by value.  Eg: actionHashMap = {worked in: 7, lived in: 10, moved to: 2} => { lived in: 10, worked in: 7, moved to: 2}. It then iterates through all the keys in the hash map for each field to come up with common combinations that aren't already in the user's history.


src/main/webapp/query-control.js
```diff
@@ -75,47 +75,102 @@ function putHistory(personType, action, location, year, userId) {
...
+  fetch(fetchUrl).then(function(response) {
+    if (!response.ok) {
+      return;
+    } else {
+      return response.json();
+    }
+  })
+    .then(function(jsonResponse) {
+      // iterate through list of recommendations
+      console.log(jsonResponse);
```
shulijones, line 150: Delete if you're done with it

src/main/java/History.java
```diff
@@ -0,0 +1,51 @@
...
+  public ArrayList<HistoryElement> getHistoryList() {
+    return historyList;
+  }
+
+  public ArrayList<HistoryElement> getHistory(String userId) {
+    // Get queries with the user's id from the database
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Filter propertyFilter = new FilterPredicate("userId", FilterOperator.EQUAL, userId);
+    Query query = new Query("historyEntity").setFilter(propertyFilter);
+      PreparedQuery results = datastore.prepare(query);
```
kyracq, line 32: Extra indent could be removed

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Set;
+
+@WebServlet("/recommendations")
+public class RecommendationsServlet extends HttpServlet {
+    private ArrayList<String> personType = new ArrayList<String>(
+      Arrays.asList("under-18", "over-18", "all-ages", "male", "female"));
+    private ArrayList<String> action = new ArrayList<String>(
+      Arrays.asList("live", "work", "moved"));
+    private ArrayList<String> location = new ArrayList<String>(
+      Arrays.asList("state","01","02","04","05","06","08","09","10","11","12","13","15","16","17","18","19","20","21","22","23","24","25","26","27","28","29","30","31","32","33","34","35","36","37","38","39","40","41","42", "44", "45", "46","47","48","49","50","51","53","54","55","56"));
```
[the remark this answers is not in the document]
kyracq: nit: also could find some way to have a loop create this array so you don't have to type out all the numbers

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // Returns a HistoryElement whose contents are completely random
+  // in order to alleviate recommendation fatigue. 
+  private HistoryElement getRandomRecommendation(String userId, ArrayList<HistoryElement> userHistory) {
+    // Max number of visualizations user could have viewed
+    int maxCombinations = 6885;
+    int counter = 0;
+    HistoryElement randomRecommendation = userHistory.get(0);
+    while (counter < maxCombinations) {
```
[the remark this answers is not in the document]
kyracq: ^^ Or you could add the invalid recommendation to a list and keep generating random queries while they're equal to one in the invalid list and the invalid list doesn't contain every query

src/main/webapp/query-control.js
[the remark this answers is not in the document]
kyracq: Just checked, seems like this is the case, resolved by only having getHistory called in loginInit, just pushed this

src/main/java/History.java
```diff
@@ -0,0 +1,51 @@
...
+        String entityPersonType = (String) entity.getProperty("personType");
+        String entityAction = (String) entity.getProperty("action");
+        String entityLocation = (String) entity.getProperty("location");
+        String entityYear = (String) entity.getProperty("year");
+        HistoryElement dataHistoryElement =
+          new HistoryElement(
+              entityUserId, entityPersonType, entityAction, entityLocation, entityYear);
+        // Check to see if it is already in the results to eliminate duplicates
+        if (!queryList.contains(dataHistoryElement)) {
+        queryList.add(dataHistoryElement);
```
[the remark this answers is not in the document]
otobin: fixed.

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
otobin: Thought abt writing a function but there are a lot of skipped numbers. Spaced them out so it looks neater.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  private void getFrequencyMaps(ArrayList<HistoryElement> userHistory) {
+    // calculate the frequencies of each type of query that the user has made
+    for (int i = 0; i < userHistory.size(); i++) {
+      HistoryElement currentElement = userHistory.get(i);
+      updateMap(personTypeMap, currentElement.getPersonType());
+      updateMap(actionMap, currentElement.getAction());
+      updateMap(locationMap, currentElement.getLocation());
+      updateMap(yearsMap, currentElement.year);
```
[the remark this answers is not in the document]
otobin: There was a bug related to year and forgot to change it back. fixed.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+    LinkedHashMap<String, Integer> reverseSortedPersonType = sortHashMapDescending(personTypeMap);
+    LinkedHashMap<String, Integer> reverseSortedAction = sortHashMapDescending(actionMap);
+    LinkedHashMap<String, Integer> reverseSortedLocation = sortHashMapDescending(locationMap);
+    LinkedHashMap<String, Integer> reverseSortedYear = sortHashMapDescending(yearsMap);
+    Set<String> personTypeKeys = reverseSortedPersonType.keySet();
+    Set<String> actionKeys = reverseSortedAction.keySet();
+    Set<String> locationKeys = reverseSortedLocation.keySet();
+    Set<String> yearKeys = reverseSortedYear.keySet();
+    ArrayList<HistoryElement> recommendationList = new ArrayList<HistoryElement>();
+    for (String personType:personTypeKeys) {
```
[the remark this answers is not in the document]
otobin: fixed

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+      for(String action:actionKeys) {
+        for (String location:locationKeys) {
+          for (String year:yearKeys) {
+            HistoryElement recommendation = new HistoryElement(userId, personType, action, location, year);
+             if (isRecommendationValid(recommendation, userHistory)) {
+               recommendationList.add(recommendation);
+               if (recommendationList.size() > 3) {
+                 return recommendationList;
```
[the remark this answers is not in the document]
otobin: The console is printing out only 5 every time it's called, so I think it's the splide that's making it repeat. Working on fixing the splide styling now

src/main/webapp/query-control.js
```diff
@@ -75,47 +75,102 @@ function putHistory(personType, action, location, year, userId) {
...
+        padding: 20,
+        pagination: false,
+      }).mount();
+    });
+}
+
+function getRecommendations() {
+  resetRecommendationList();
+  const recommendationList = document.getElementById('recommendation-list');
+  const fetchUrl = '/recommendations?user-id=' + getUserId();
```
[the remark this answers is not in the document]
otobin: If the user currently doesn't have any history the recommendations shows them the most general recommendations that would be applicable to most

src/main/webapp/query-control.js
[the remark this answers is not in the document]
otobin: got it! Will pull from your branch

src/main/webapp/query-control.js
```diff
@@ -75,47 +75,102 @@ function putHistory(personType, action, location, year, userId) {
...
+  })
+    .then(function(jsonResponse) {
+      // iterate through list of recommendations
+      console.log(jsonResponse);
+      jsonResponse.forEach((historyElement) => {
+        if (historyElement === null) return;
+        addHistoryToPage(historyElement, recommendationList);
+      });
+      new Splide('.splide2', {
+        type: 'loop',
```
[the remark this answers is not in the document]
otobin: nice! will adjust

src/main/java/History.java
otobin: true! fixed.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+    // Max number of visualizations user could have viewed
+    int maxCombinations = 6885;
+    int counter = 0;
+    HistoryElement randomRecommendation = userHistory.get(0);
+    while (counter < maxCombinations) {
+      String randomPersonType = personType.get(getRandomInt(personType.size()));
+      String randomAction = action.get(getRandomInt(action.size()));
+      String randomLocation = location.get(getRandomInt(location.size()));
+      String randomYear = years.get(getRandomInt(years.size()));
+      randomRecommendation = new HistoryElement(
```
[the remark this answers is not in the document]
otobin: true. Visualization element maybe?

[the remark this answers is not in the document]
otobin: changed to VisualizationData

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // Return suggested recommendations based on the history. Create hashmaps for each input for all of the queries,
+  // then create four combinations based on the most frequented fields.
+  private ArrayList<HistoryElement> getRecommendations(ArrayList<HistoryElement> userHistory, String userId) {
+    getFrequencyMaps(userHistory);
+    LinkedHashMap<String, Integer> reverseSortedPersonType = sortHashMapDescending(personTypeMap);
+    LinkedHashMap<String, Integer> reverseSortedAction = sortHashMapDescending(actionMap);
+    LinkedHashMap<String, Integer> reverseSortedLocation = sortHashMapDescending(locationMap);
+    LinkedHashMap<String, Integer> reverseSortedYear = sortHashMapDescending(yearsMap);
```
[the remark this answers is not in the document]
otobin: I'm sorting the map by value here, such that actions that the user researches the most are at the front of the list.
Eg: {worked in: 7, lived in: 10, moved to: 2} => { lived in: 10, worked in: 7, moved to: 2}. That way, when I iterate through the lists with all of the query fields in lines 136-150 I would get combinations that the user is likely to make. I think if I sorted only the keyset I would only get it in alphabetical order, and if I sorted only the values I would have the sorted list but I wouldn't have a way to access the keys from only the value in the map interface. You bring up a good point that the keyset may not retain the sorted order, but after researching/debugging I believe it does https://stackoverflow.com/questions/2923856/is-the-order-guaranteed-for-the-return-of-keys-and-values-from-a-linkedhashmap-o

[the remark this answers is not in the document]
shulijones: Ah, I understand now - sorry, I definitely hadn't quite figured out what was happening here. That makes sense to me, and I agree that it sounds like the keySet does retain the order. So I'll resolve this.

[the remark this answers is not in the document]
otobin: It's all good! I will add more commenting so it's clear

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
otobin: did the later (I think?)

src/main/webapp/query-control.js
otobin: facts

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  // uses java.util.random to get a random integer between 
+  // 0 and upperBound-1 in order to access a random element 
+  // for each of the options
+  private int getRandomInt(int upperBound) {
+    Random randomObject = new Random();
+    int randomInt = randomObject.nextInt(upperBound);
+    return randomInt;
+  }
+
+  // CHeck to see if the current recommendation is valid by making sure that it isn't an impossible combination
```
shulijones, line 61: nittest nit: uncapitalize H

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+      VisualizationData randomRecommendation = new VisualizationData(
+          userId, randomPersonType, randomAction, randomLocation, randomYear);
+      if (isRecommendationValid(randomRecommendation, userHistory) && !invalidRecommendations.contains(randomRecommendation)) {
+        return randomRecommendation;
+      } else {
+        invalidRecommendations.add(randomRecommendation);
+      }
+      // If the user has looked at every visualization, return the first in the history.
+      if (invalidRecommendations.size() >= 6885) {
+        isValid = true;
```
shulijones, line 88: It looks like you're not using isValid to keep track of validity at all, since when you find a valid recommendation, you just return it. So it might make more sense with a different name -- or, you could just run your while loop on while (invalidRecommendations.size() < 6885).

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+      String randomPersonType = personType.get(getRandomInt(personType.size()));
+      String randomAction = action.get(getRandomInt(action.size()));
+      String randomLocation = location.get(getRandomInt(location.size()));
+      String randomYear = years.get(getRandomInt(years.size()));
+      VisualizationData randomRecommendation = new VisualizationData(
+          userId, randomPersonType, randomAction, randomLocation, randomYear);
+      if (isRecommendationValid(randomRecommendation, userHistory) && !invalidRecommendations.contains(randomRecommendation)) {
+        return randomRecommendation;
+      } else {
+        invalidRecommendations.add(randomRecommendation);
```
shulijones, line 84: Maybe name this unuseableRecommendations instead? Since it's a collection of recommendations that are either invalid or already seen. That would also make your if statement (line 81) make more sense, since right now it reads like you're checking if it's invalid twice

src/main/webapp/query-control.js
```diff
@@ -109,13 +109,21 @@ function resetHistoryList() {
 
 function toggleHistory(shouldShow) {
```
shulijones, line 110: Maybe rename this function to toggleLoggedInContent or something that indicates it's toggling all the things that the person has to be logged in to see?

src/main/webapp/query-control.js
```diff
@@ -137,9 +145,34 @@ function getHistory() {
...
+      return response.json();
+    }
+  })
+    .then(function(jsonResponse) {
+      // iterate through list of recommendations
+      jsonResponse.forEach((historyElement) => {
+        if (historyElement === null) return;
+        addHistoryToPage(historyElement, recommendationList);
+      });
+      new Splide('#splide2', {
```
shulijones, line 174: I kind of wish there was a way to tell it to only show arrows if there are more items that aren't visible on the page. My screen can fit up to 6 recommendations, so I never need the arrows, I can always see them all. But if someone else is on phone, they might still want them. IDK if this is possible, though.

src/main/webapp/style.css
```diff
@@ -365,6 +365,41 @@ table {
...
+  margin: 0 auto 15px auto;
+  width: 150px;
+}
+
+#recommendation-list li {
+  border: 1px solid #e7e7e9;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+  box-shadow: 4px 4px 8px #e7e7e9;
```
shulijones, line 386: this should be up with the b's for alphabetical sorting

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+      counter += 1;
+    }
+    // If the user has looked at EVERy visualization (not likely), just return the first 
+    // one in the history. 
+    return randomRecommendation;
+  }
+
+  // Takes in a map and updates the key/value pairs
+  private void updateMap(HashMap<String, Integer> frequencyMap, String item) {
```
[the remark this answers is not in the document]
shulijones: bump

src/main/java/History.java
```diff
@@ -0,0 +1,51 @@
+package com.google.sps.data;
+import com.google.sps.data.VisualizationData;
```
kyracq, line 2: I think this can be removed since it's in the same package as this class

src/main/java/History.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import java.util.ArrayList;
+import java.util.List;
```
kyracq, line 12: Doesn't look like you're using this so could remove this line

src/main/java/History.java
```diff
@@ -0,0 +1,51 @@
...
+        String entityLocation = (String) entity.getProperty("location");
+        String entityYear = (String) entity.getProperty("year");
+        VisualizationData dataHistoryElement =
+          new VisualizationData(
+              entityUserId, entityPersonType, entityAction, entityLocation, entityYear);
+        // Check to see if it is already in the results to eliminate duplicates
+        if (!queryList.contains(dataHistoryElement)) {
+          queryList.add(dataHistoryElement);
+        }
+      }
```
kyracq, line 48: Looks like this block is indented too far in

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  private int getRandomInt(int upperBound) {
+    Random randomObject = new Random();
+    int randomInt = randomObject.nextInt(upperBound);
+    return randomInt;
+  }
+
+  // CHeck to see if the current recommendation is valid by making sure that it isn't an impossible combination 
+  // and isn't already one that the user has viewed.
+  private boolean isRecommendationValid(VisualizationData recommendation, ArrayList<VisualizationData> userHistory) {
+      return (!(userHistory.contains(recommendation) ||
```
kyracq, line 64: Extra indents here

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+    // Iterate through all fields and find combinations that the user hasn't made yet
+    // and return the top 3.
+    for (String personType : personTypeKeys) {
+      for(String action : actionKeys) {
+        for (String location : locationKeys) {
+          for (String year : yearKeys) {
+            VisualizationData recommendation = new VisualizationData(userId, personType, action, location, year);
+             if (isRecommendationValid(recommendation, userHistory)) {
+               recommendationList.add(recommendation);
+               if (recommendationList.size() > 3) {
```
kyracq, line 149: Comment on 141 says you're returning top 3, but I think this goes up to 4

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+        frequencyMap.entrySet().stream().sorted(Map.Entry.comparingByValue(Comparator.reverseOrder()))
+                .forEachOrdered(x -> reverseSortedMap.put(x.getKey(), x.getValue()));
+    return reverseSortedMap;
+  }
+
+  // Return suggested recommendations based on the history. Create hashmaps for each input for all of the queries,
+  // then create four combinations based on the most frequented fields.
+  private ArrayList<VisualizationData> getRecommendations(ArrayList<VisualizationData> userHistory, String userId) {
+    // Populate the frequency maps for each field
+    getFrequencyMaps(userHistory);
```
kyracq, line 127: Not sure exactly where this issue comes from or if this is the expected behavior but the recommendations based on history seem to be exclusively all ages population in each U.S. state for various years. I thought it might've been because there were too many queries already in my history, so I cleared my history and made four queries all with "worked in", but the recommendation were the same. 
image

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userId = request.getParameter("user-id");
+    History userHistoryObj = new History(userId);
+    ArrayList<VisualizationData> userHistory = userHistoryObj.getHistoryList();
+    if (userHistory.isEmpty()) {
+      String json = new Gson().toJson(getDefaultRecommendations(userId));
+      response.getWriter().write(json);
+    } else {
+      ArrayList<VisualizationData> recommendations = getRecommendations(userHistory, userId);
+      // If the recommendation algorithm is unable to generate more than 4 possibilities
```
kyracq, line 170: Adding to previous comment, I seem to always get four from the getRecommendations function and 1 random

src/main/webapp/query-control.js
kyracq: Could check window width right above this (window.innerWidth) and turn off arrows past a certain width (by adding "arrows: false" to the splide options)

src/main/webapp/style.css
```diff
@@ -365,6 +365,41 @@ table {
   padding-left: 35px;
 }
 
+#recommendation-list a {
```
kyracq, line 368: Since recommendation list has the same styling as history list, you could remove the CSS for one of the two and rename it with a class called something like "visualization thumbnails" and add the class to both

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+    LinkedHashMap<String, Integer> reverseSortedLocation = sortHashMapDescending(locationMap);
+    LinkedHashMap<String, Integer> reverseSortedYear = sortHashMapDescending(yearsMap);
+    // Get the keyset from each frequency map with the guaranteed order from sorting
+    Set<String> personTypeKeys = reverseSortedPersonType.keySet();
+    Set<String> actionKeys = reverseSortedAction.keySet();
+    Set<String> locationKeys = reverseSortedLocation.keySet();
+    Set<String> yearKeys = reverseSortedYear.keySet();
+    ArrayList<VisualizationData> recommendationList = new ArrayList<VisualizationData>();
+    // Iterate through all fields and find combinations that the user hasn't made yet
+    // and return the top 3.
```
nsri-google, line 141: shouldn't this be top 4?

src/main/java/RecommendationsServlet.java
nsri-google, line 149: you can make the number of recommendations be a parameter and you can return the requested number of recommendations and padding with random recommendations if necessary.

src/main/java/RecommendationsServlet.java
otobin: yeet

src/main/webapp/query-control.js
otobin: changed to toggleUserContent()

src/main/java/RecommendationsServlet.java
otobin: fixed.

src/main/java/RecommendationsServlet.java
otobin: true

src/main/webapp/style.css
otobin: fixed.

src/main/java/RecommendationsServlet.java
otobin: Hmm that is weird. I didn't run into that issue. Maybe if you try refreshing the page? Or mvn clean?

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+    return defaultRecommendations;
+  }
+
+  // uses java.util.random to get a random integer between 
+  // 0 and upperBound-1 in order to access a random element 
+  // for each of the options
+  private int getRandomInt(int upperBound) {
+    Random randomObject = new Random();
+    int randomInt = randomObject.nextInt(upperBound);
+    return randomInt;
```
anicr7, line 58: Can be one line - return new Random().nextInt(upperBound);

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Random;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+@WebServlet("/recommendations")
+public class RecommendationsServlet extends HttpServlet {
+  private ArrayList<String> personType = new ArrayList<String>(
```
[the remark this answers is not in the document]
anicr7: Agree, changing it in multiple places will be a painful and will lead to breakages.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  // for each of the options
+  private int getRandomInt(int upperBound) {
+    Random randomObject = new Random();
+    int randomInt = randomObject.nextInt(upperBound);
+    return randomInt;
+  }
+
+  // CHeck to see if the current recommendation is valid by making sure that it isn't an impossible combination 
+  // and isn't already one that the user has viewed.
+  private boolean isRecommendationValid(VisualizationData recommendation, ArrayList<VisualizationData> userHistory) {
```
anicr7, line 63: This should be a hashmap instead.  That will take care of the duplications by default.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+      String randomYear = years.get(getRandomInt(years.size()));
+      VisualizationData randomRecommendation = new VisualizationData(
+          userId, randomPersonType, randomAction, randomLocation, randomYear);
+      if (isRecommendationValid(randomRecommendation, userHistory) && !invalidRecommendations.contains(randomRecommendation)) {
+        return randomRecommendation;
+      } else {
+        invalidRecommendations.add(randomRecommendation);
+      }
+      // If the user has looked at every visualization, return the first in the history.
+      if (invalidRecommendations.size() >= 6885) {
```
anicr7, line 87: I would avoid using magic numbers such as this that can easily be broken.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+        (recommendation.getPersonType() == "under-18" && recommendation.getAction() == "work") ||
+        (recommendation.getPersonType() == "under-18" && recommendation.getAction() == "moved"))); 
+  }
+
+  // Returns a VisualizationData whose contents are completely random
+  // in order to alleviate recommendation fatigue. 
+  private VisualizationData getRandomRecommendation(String userId, ArrayList<VisualizationData> userHistory) {
+    ArrayList<VisualizationData> invalidRecommendations = new ArrayList<VisualizationData>();
+    boolean isValid = false;
+    while (!isValid) {
```
anicr7, line 74: You can be stuck in this loop based on how random the integer generation actually is. This might be ok for now, but a cleaner solution might be to generate only valid combinations.

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
anicr7: There is a even more simpler one-liner

frequencyMap.merge(item, 1, Integer::sum);

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  // Takes in a map and updates the key/value pairs
+  private void updateMap(HashMap<String, Integer> frequencyMap, String item) {
+    if (frequencyMap.containsKey(item)) {
+      frequencyMap.put(item, frequencyMap.get(item) + 1);
+    } else {
+      frequencyMap.put(item, 1);
+    }
+  }
+
+  private void getFrequencyMaps(ArrayList<VisualizationData> userHistory) {
```
anicr7, line 103: Maybe this should be called updateFrequencyMaps as I don't see any gets being done here.

src/main/java/History.java
otobin: I thought you had to import it but you're right, it compiles no problem.

src/main/java/History.java
otobin: fixed

src/main/java/History.java
otobin: nice catch! fixed

src/main/java/RecommendationsServlet.java
otobin: fixed.

src/main/java/RecommendationsServlet.java
otobin: Yes, it should be four. Adjusted comment.

src/main/java/RecommendationsServlet.java
otobin: Yes, for the most part you should always get four from getRecommendations and 1 random except if you have many queries with only one field changed. Eg: "How many people lived in each California County in 2018" and "How many people lived in each Florida County in 2018". With these two elements in the history, there are no new combinations we can make, so more random ones are returned. Ideally the algorithm would return more visualizations about the existing states, which I'll implement if I have time.

src/main/webapp/query-control.js
otobin: Fixed so that the user should only see arrows if their window is less than 1250 px

src/main/webapp/style.css
otobin: I tried this but it ended up erasing the styling for the images like the border and link styling.

src/main/java/RecommendationsServlet.java
otobin: yes. Fixed.

src/main/java/RecommendationsServlet.java
otobin: fixed.

src/main/java/RecommendationsServlet.java
otobin: Fixed so that maxVisualizations is the size of all arraylists multiplied together.

src/main/java/RecommendationsServlet.java
otobin: The userHistory should be a hashMap?

src/main/java/RecommendationsServlet.java
otobin: Changed to updateFrequencyMaps

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
otobin: Used map.merge(key, value, Int::sum) instead

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+    while (!isValid) {
+      String randomPersonType = personType.get(getRandomInt(personType.size()));
+      String randomAction = action.get(getRandomInt(action.size()));
+      String randomLocation = location.get(getRandomInt(location.size()));
+      String randomYear = years.get(getRandomInt(years.size()));
+      VisualizationData randomRecommendation = new VisualizationData(
+          userId, randomPersonType, randomAction, randomLocation, randomYear);
+      if (isRecommendationValid(randomRecommendation, userHistory) && !invalidRecommendations.contains(randomRecommendation)) {
+        return randomRecommendation;
+      } else {
```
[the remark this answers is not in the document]
otobin: fixed.

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
otobin: Changed 6885 to an integer with all of the sizes multiplied together.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userId = request.getParameter("user-id");
+    History userHistoryObj = new History(userId);
+    ArrayList<VisualizationData> userHistory = userHistoryObj.getHistoryList();
+    if (userHistory.isEmpty()) {
+      String json = new Gson().toJson(getDefaultRecommendations(userId));
+      response.getWriter().write(json);
```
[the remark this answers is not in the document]
otobin: Updated function so there's only one response.getWriter().write(json)

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+      ArrayList<VisualizationData> recommendations = getRecommendations(userHistory, userId);
+      // If the recommendation algorithm is unable to generate more than 4 possibilities
+      // which happens when the user makes very similar searches, populate the rest with random elements.
+      if (recommendations.size() < 4) {
+        for (int i = 0; i < 4 - recommendations.size(); i++) {
+          recommendations.add(getRandomRecommendation(userId, userHistory));
+        }
+      }
+      VisualizationData randomRecommendation = getRandomRecommendation(userId, userHistory);
+      recommendations.add(randomRecommendation);
```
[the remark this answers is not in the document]
otobin: fixed.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+    History userHistoryObj = new History(userId);
+    ArrayList<VisualizationData> userHistory = userHistoryObj.getHistoryList();
+    if (userHistory.isEmpty()) {
+      String json = new Gson().toJson(getDefaultRecommendations(userId));
+      response.getWriter().write(json);
+    } else {
+      ArrayList<VisualizationData> recommendations = getRecommendations(userHistory, userId);
+      // If the recommendation algorithm is unable to generate more than 4 possibilities
+      // which happens when the user makes very similar searches, populate the rest with random elements.
+      if (recommendations.size() < 4) {
```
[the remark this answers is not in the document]
otobin: fixed so that getRecommendations should always return four recommendations.

src/main/java/RecommendationsServlet.java
```diff
@@ -0,0 +1,184 @@
...
+        for (int i = 0; i < 4 - recommendations.size(); i++) {
+          recommendations.add(getRandomRecommendation(userId, userHistory));
+        }
+      }
+      VisualizationData randomRecommendation = getRandomRecommendation(userId, userHistory);
```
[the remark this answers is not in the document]
otobin: Previously, getRecommendations() returned 4 recommendations based on history, so adding one more random recommendation would add the 5th. I've updated the algorithm so that getRandomRecommendations now returns all 5 for clarity.

src/main/java/RecommendationsServlet.java
otobin: The three of us decided that we should only show five at a time, so I've updated getRecommendations to only return five.

src/main/java/RecommendationsServlet.java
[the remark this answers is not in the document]
otobin: I also agree that this is very bug prone. I'm not sure where the best place would be. Should I add a function that gets the data from the maps in the query servlet? Or should the query servlet get them from here?

src/main/java/RecommendationsServlet.java
otobin: Updated so the function iterates as long as the unuseableRecommendation array is not full of all possible recommendations. If a valid recommendation is generated, then it is returned immediately.

I don't think this fixes the problem bc invalid recommendations can still be generated

src/main/java/RecommendationsServlet.java
kyracq: Yeah after clearing everything it's working great! Thanks

src/main/webapp/style.css
kyracq: Oh I guess there'd be a few other things you'd need to change, so it can be kept the way it is, thanks

src/main/java/RecommendationsServlet.java
kyracq: Oh I see, thanks!

PR #59, opened 2020-08-26
Dynamic dropdowns

shulijones:
(not yet deployed)
When a user picks one dropdown option, exclude others that are not relevant (e.g. if they pick 'children', exclude 'work').

### DIFF
--- a/src/main/webapp/query-control.js
+++ b/src/main/webapp/query-control.js
@@ -81,7 +81,7 @@ function getHistory() {
   const historyContainer = document.getElementById('history');
   historyContainer.innerHTML = '';
   const header = document.createElement('p');
-  header.innerText = "Pages you've Viewed";
+  header.innerText = 'Pages you\'ve Viewed';
   historyContainer.appendChild(header);
   const fetchUrl = '/history?user-id=' + getUserId();
   fetch(fetchUrl).then(function(response) {
@@ -291,7 +291,7 @@ function replaceValueIfEmpty(dataListId) {
 }
 
 // When the user picks certain dropdown values,
-// exclude the combinations that don't have data available 
+// exclude the combinations that don't have data available
 const dropdownsToExclude = new Map();
 dropdownsToExclude.set(
       'worked in', ['children'],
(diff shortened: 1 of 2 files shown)


src/main/webapp/query-control.js
```diff
@@ -290,6 +290,52 @@ function replaceValueIfEmpty(dataListId) {
...
+    options = defaultDropdowns.get('person-type');
+  } else {
+    dropdown = document.getElementById('action');
+    options = defaultDropdowns.get('action');
+  }
+  dropdown.innerHTML = '';
+  const currentValue = document.getElementById(type + '-list').value;
+
+  options.forEach((option) => {
+    if (!(dropdownsToExclude.has(currentValue) &&
```
kyracq, line 331: I think a comment explaining this if statement would be helpful

shulijones: Sure, good point! Added.

src/main/webapp/query-control.js
```diff
@@ -290,6 +290,56 @@ function replaceValueIfEmpty(dataListId) {
   }
 }
 
+// When the user picks certain dropdown values,
+// exclude the combinations that don't have data available
+const dropdownsToExclude = new Map();
+dropdownsToExclude.set(
```
nsri-google, line 296: the same values have to be listed twice once for persontype to action and once again for action to persontype. You could instead make this just a set of personType+action. Then when checking you compute personType+action and check if it exists in the set.

src/main/webapp/query-control.js
```diff
@@ -290,6 +290,56 @@ function replaceValueIfEmpty(dataListId) {
...
+  }
+  dropdown.innerHTML = '';
+  const currentValue = document.getElementById(type + '-list').value;
+
+  // Iterate through all potential options
+  options.forEach((option) => {
+    // Check if the current option is in the exclusions list for the value that
+    // the user entered; if it's not, add it to the dropdown
+    if (!(dropdownsToExclude.has(currentValue) &&
+        dropdownsToExclude.get(currentValue).includes(option[0]))) {
```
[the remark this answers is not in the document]
shulijones: Good point. Done

src/main/webapp/query-control.js
shulijones: Ah, good idea. I've done this, hopefully in a way that's not too clunky.

PR #62, opened 2020-08-28
Graph

kyracq:
- Line graph showing change in population over time when you query state data.
- County dropdown to see yearly data for a county in queried state.

### DIFF
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -273,8 +273,9 @@ public class QueryServlet extends HttpServlet {
       }
 
       String formattedData;
+      boolean isCountyQuery = (!location.equals("state") && !stateTotal) && !singleCounty;
       try {
-        formattedData = DataFormatter.reformatDataArray(data, dataRow, !location.equals("state"));
+        formattedData = DataFormatter.reformatDataArray(data, dataRow, isCountyQuery);
       } catch (InvalidObjectException e) {
         response.setStatus(HttpServletResponse.SC_BAD_GATEWAY);
         response
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -128,13 +128,18 @@ public class QueryServlet extends HttpServlet {
         + dataRow.substring(0, (dataRow.contains("_") ? dataRow.indexOf("_") : dataRow.length()));
   }
 
+  // Return stirng representation of a JSON object containing
+  // an errorMessage
   private String sendError(String errorMessage) {
     JsonObject jsonResponse = new JsonObject();
     jsonResponse.addProperty("errorMessage", errorMessage);
     return jsonResponse.toString();
   }
 
-  private String getLocation(
+  // Return the string to pass in the "for" (location) parameter to
+  // the Census API based on whether the query is for a single state's
+  // total population, a single county, or multiple counties or states. 
+  private String getLocationQueryString(
       boolean stateTotal, boolean singleCounty, String location, String county) {
     if (stateTotal) {
       return "state:" + location;
@@ -238,7 +243,7 @@ public class QueryServlet extends HttpServlet {
                 + "?get=NAME,"
                 + dataRow
                 + "&for="
-                + getLocation(stateTotal, singleCounty, location, county)
+                + getLocationQueryString(stateTotal, singleCounty, location, county)
                 + "&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
     String censusTableLink = getCensusTableLink(dataRow, dataTablePrefix, yearStr);
 
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -135,7 +135,7 @@ public class QueryServlet extends HttpServlet {
   }
 
   private String getLocation(
-    boolean stateTotal, boolean singleCounty, String location, String county) {
+      boolean stateTotal, boolean singleCounty, String location, String county) {
     if (stateTotal) {
       return "state:" + location;
     } else if (singleCounty) {
--- a/src/main/java/QueryServlet.java
+++ b/src/main/java/QueryServlet.java
@@ -134,6 +134,19 @@ public class QueryServlet extends HttpServlet {
     return jsonResponse.toString();
   }
 
+  private String getLocation(boolean stateTotal, boolean singleCounty,
+      String location, String county) {
+    if (stateTotal) {
+      return "state:" + location;
+    } else if (singleCounty) {
+      return "county:" + county + "&in=state:" + location;
+    } else if (location.equals("state")) {
+      return "state:*";
+    } else {
+      return "county:*&in=state:" + location;
+    }
+  }
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String personType = request.getParameter("person-type");
@@ -141,17 +154,29 @@ public class QueryServlet extends HttpServlet {
     String location = request.getParameter("location");
     String yearStr = request.getParameter("year");
     int year = Integer.parseInt(yearStr);
+    String stateTotalString = request.getParameter("state-total");
+    String county = request.getParameter("county");
+    boolean stateTotal = false;
+    boolean singleCounty = false;
+    if (stateTotalString != null) {
+      stateTotal = stateTotalString.equals("true");
+    }
+    if (county != null) {
+      singleCounty = true;
+    }
 
     response.setContentType("application/json;");
-
-    String query = personType + action + location + yearStr;
-    CensusData cachedData = ofy().load().type(CensusData.class).id(query).now();
-    if (cachedData != null) {
-      JsonObject jsonResponse = new JsonObject();
-      jsonResponse.addProperty("censusData", cachedData.getData());
-      jsonResponse.addProperty("tableLink", cachedData.getTableLink());
-      response.getWriter().println(jsonResponse.toString());
-      return;
+    String query = null;
+    if (!stateTotal && !singleCounty) {
+      query = personType + action + location + yearStr;
+      CensusData cachedData = ofy().load().type(CensusData.class).id(query).now();
+      if (cachedData != null) {
+        JsonObject jsonResponse = new JsonObject();
+        jsonResponse.addProperty("censusData", cachedData.getData());
+        jsonResponse.addProperty("tableLink", cachedData.getTableLink());
+        response.getWriter().println(jsonResponse.toString());
+        return;
+      }
     }
 
     if (!queryToDataRowGeneric.containsKey(action)) {
@@ -213,9 +238,8 @@ public class QueryServlet extends HttpServlet {
                 + "?get=NAME,"
                 + dataRow
                 + "&for="
-                + (location.equals("state") ? "state:*" : "county:*&in=state:" + location)
+                + getLocation(stateTotal, singleCounty, location, county)
                 + "&key=ea65020114ffc1e71e760341a0285f99e73eabbc");
-
     String censusTableLink = getCensusTableLink(dataRow, dataTablePrefix, yearStr);
 
     HttpURLConnection connection = (HttpURLConnection) fetchUrl.openConnection();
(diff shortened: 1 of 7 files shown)


src/main/java/QueryServlet.java
```diff
@@ -134,24 +134,49 @@ private String sendError(String errorMessage) {
     return jsonResponse.toString();
   }
 
+  private String getLocation(
```
shulijones, line 137: Could you update the name of this method to make it more clear what it does? In particular, it's confusing to me that getLocation requires an input parameter of location.

shulijones: A comment above the method would be great too actually

src/main/webapp/query-control.js
```diff
@@ -31,6 +31,11 @@ function clearPreviousResult() {
   document.getElementById('census-link').style.display = 'none';
   document.getElementById('toggle-data-btn').style.display = 'none';
   document.getElementById('map-options').style.display = 'none';
+  document.getElementById('yearly-total-chart').style.display = 'none';
```
shulijones, line 34: This function is now super long. Do we want to switch to having a list of ID names, and we do a forEach on the list to set the display to none? (Plus a couple other actions that are standalone, of course.)

src/main/webapp/index.html
```diff
@@ -183,6 +184,18 @@ <h1><a href="/">Census Viz</a></h1>
       <a id="census-link" target="_blank" rel="noopener">Explore more data from the Census API</a>
       <br>
       <div id="data-table"></div>
+      <div id="yearly-total-chart-loading-msg"></div>
+      <div id="yearly-total-chart" class="yearly-chart"></div>
+      <div id="yearly-county-chart-loading-msg"></div>
+      <form id="county-form" onsubmit="createYearlyCountyChart(); return false;">
+        <label for="county-list">View yearly data for a county</label>
```
shulijones, line 191: What would you think of phrasing this as "View yearly data for" ? Fitting with our main query that uses the dropdown as kind of part of the sentence.
Alternatively, if that's too weird, I think it'd be nice to add a colon ("...for a county:") to separate the text and the label a bit more.

src/main/webapp/data-visualization.js
```diff
@@ -23,6 +23,20 @@ async function getGeoData(locationInfo, isCountyQuery) {
   }
 }
 
+// Create dropdown menu of couties in queried state
```
shulijones, line 26: nit: typo in counties

src/main/webapp/data-visualization.js
```diff
@@ -508,6 +529,142 @@ function toggleDataTable() {
...
+  const dateAxis = chart.xAxes.push(new am4charts.DateAxis());
+  dateAxis.dataFields.category = 'year';
+  dateAxis.title.text = 'Year';
+  dateAxis.baseInterval = {timeUnit: 'year'};
+  const valueAxis = chart.yAxes.push(new am4charts.ValueAxis());
+  valueAxis.title.text = description;
+  const series = chart.series.push(new am4charts.LineSeries());
+  series.dataFields.valueY = 'value';
+  series.dataFields.dateX = 'year';
+  series.name = description;
```
shulijones, line 544: I'm not sure exactly where it is, but I think this is the line that adds the "legend"? I feel like it could be deleted -- since there's just one line on the graph, it's not that relevant or helpful to have something that just matches that line to the title of the graph.

src/main/webapp/data-visualization.js
```diff
@@ -508,6 +529,142 @@ function toggleDataTable() {
...
+  const color = getColor();
+  series.stroke = am4core.color(color);
+  series.yAxis = valueAxis;
+
+  const bullet = series.bullets.push(new am4charts.CircleBullet());
+  bullet.stroke = am4core.color('#fff');
+  bullet.fill = am4core.color(color);
+  bullet.strokeWidth = 2;
+  bullet.tooltipText = '{valueY}';
+  bullet.circle.radius = 4;
```
shulijones, line 558: I think it would be nice if the bullets were a bit bigger. They would stand out more and it would also be easier to trigger the tooltip text (right now I sometimes mouse past them by accident instead of over them).

src/main/webapp/data-visualization.js
```diff
@@ -508,6 +529,142 @@ function toggleDataTable() {
   }
 }
 
+function displayYearlyChart(chartId, data, description, location) {
+  const chart = am4core.create(chartId, am4charts.XYChart);
+  chart.data = data;
+  const dateAxis = chart.xAxes.push(new am4charts.DateAxis());
+  dateAxis.dataFields.category = 'year';
+  dateAxis.title.text = 'Year';
+  dateAxis.baseInterval = {timeUnit: 'year'};
```
shulijones, line 538: I'm not sure where to bring this up, but I'd like if the charts had a bit more height. They're very wide and short, and I think it'd be interesting to stretch out the y-axis a little more (maybe that could be done around these lines of code?) and see how that looks.

src/main/webapp/data-visualization.js
```diff
@@ -508,6 +529,142 @@ function toggleDataTable() {
...
+  for (let year = 2010; year <= 2018; year++) {
+    await fetch(getFetchUrl('query', personType, action, location, year) +
+      '&state-total=true')
+      .then((response) => response.json().then((jsonResponse) => ({
+        data: jsonResponse,
+        success: response.ok,
+        status: response.status,
+      })))
+      .then((response) => {
+        if (response.success) {
```
shulijones, line 590: If response isn't success, you should display an error

src/main/webapp/data-visualization.js
```diff
@@ -508,6 +529,142 @@ function toggleDataTable() {
...
+  const locationDropdown = document.querySelector(
+    '#location option[data-value=\'' + query.location + '\']');
+  let locationInfo;
+  let stateNumber;
+  if (locationDropdown !== null &&
+    !locationDropdown.classList.contains('autocomplete-item')) {
+    // User picked a location from the dropdown
+    stateNumber = query.location;
+  } else { // Have to manually find which state the location is in
+    locationInfo = await findStateOfLocation(query.location);
```
shulijones, line 633: this looks similar to some other code in data-viz (which I only remember because I wrote it). Maybe it would make sense to turn it into a function instead of having it written out twice?

src/main/webapp/index.html
```diff
@@ -183,6 +184,18 @@ <h1><a href="/">Census Viz</a></h1>
...
+        <input type="list" list="county" name="county" id="county-list" autocomplete="off" required
+          onfocus="storeValueAndEmpty('county')" onfocusout="replaceValueIfEmpty('county')" value="Select a county">
+        <datalist id="county"></datalist>
+        </input>
+        <input type="submit" value="Submit">
```
shulijones, line 196: Any reason why this doesn't say "Go" like our main query does?

src/main/webapp/index.html
```diff
@@ -183,6 +184,18 @@ <h1><a href="/">Census Viz</a></h1>
       <a id="census-link" target="_blank" rel="noopener">Explore more data from the Census API</a>
       <br>
       <div id="data-table"></div>
+      <div id="yearly-total-chart-loading-msg"></div>
```
shulijones, line 187: It'd be nice to have an h3 here, like we do for the history and recommendations, saying like "Year-by-year data" or something.

shulijones (commented):
This looks awesome! Thank you for putting it all together. I left a bunch of comments but nothing huge.

src/main/webapp/data-visualization.js
```diff
@@ -23,6 +23,20 @@ async function getGeoData(locationInfo, isCountyQuery) {
...
+// Create dropdown menu of couties in queried state
+// for yearly population change in county chart.
+function createCountyDropdown(geoDataFeatures) {
+  const datalist = document.getElementById('county');
+  geoDataFeatures = geoDataFeatures.reverse(); // Make dropdown alphabetical
+  datalist.innerHTML = '';
+  for (let i = 0; i < geoDataFeatures.length; i++) {
+    const optionElem = document.createElement('option');
+    optionElem.value = geoDataFeatures[i].properties.name;
+    optionElem.setAttribute('data-value', geoDataFeatures[i].id.slice(2, 5));
```
otobin, line 35: why (2,5)?

src/main/webapp/data-visualization.js
```diff
@@ -298,18 +314,20 @@ function getMapsData(censusDataArray) {
     // ^^ ["Contra Costa County", "California"]
     // Trim space and region suffix
     const countyRegex = new RegExp('( County| City and Borough|' +
-        ' Borough| Municipality| Census Area| city| City)$');
+      ' Borough| Municipality| Census Area| city| City)$');
```
otobin, line 317: nit: add Parish

otobin (approved):
Looks so good! Left just two nit pick comments

src/main/webapp/data-visualization.js
kyracq: Right now I show a "no data available" message if there is no data returned successfully for 2010-2018. If I add an error message on not success, it'd keep showing for all of the years, but also show the graph with data for the successful years. So, I was thinking the "no data available" message when all years are not success would be the most useful to the user. Let me know what you'd had in mind for error messages

src/main/java/QueryServlet.java
kyracq: Done!

src/main/webapp/query-control.js
kyracq: Great idea! Done!

src/main/webapp/index.html
kyracq: Switched to a question format to try to match main query: "how many {personType} {action} in {county} each year?", let me know your thoughts on this or whether you prefer "yearly data for"

src/main/webapp/data-visualization.js
kyracq: Fixed!

src/main/webapp/data-visualization.js
kyracq: Totally agree, legend removed

src/main/webapp/data-visualization.js
kyracq: Good point! Changed radius to 8, let me know if it's still too small

src/main/webapp/data-visualization.js
kyracq: Changed height from 300px to 500px in CSS

src/main/webapp/data-visualization.js
kyracq: Done!

src/main/webapp/index.html
kyracq: Nope, switched to "Go"

src/main/webapp/index.html
kyracq: Agreed! Added this header

src/main/webapp/data-visualization.js
kyracq: Added a comment about this, it's to get last 3 digits of county id which is county number without the state number

src/main/webapp/data-visualization.js
kyracq: Fixed!

src/main/webapp/index.html
shulijones: I like that! Thanks.

src/main/webapp/data-visualization.js
shulijones: LGTM

src/main/webapp/data-visualization.js
shulijones: I see that you do that for the individual county graphs, but it looks like you don't for the state graph? I agree that no data available makes sense.

src/main/webapp/data-visualization.js
shulijones: I think this is good.

src/main/webapp/data-visualization.js
kyracq: Ah, you're right, added error message for state charts

nsri-google (approved):

anicr7 (approved):

PR #63, opened 2020-08-28
Reorganize top buttons

shulijones:
Rearrange the buttons above the visualization so that they form a clean line, and update the css and javascript to treat them all together to reduce code complexity.

I think this is a big improvement on the vertical stack of buttons we had previously, and they appear and disappear more cleanly too because they're all grouped together. This is primarily a visual improvement to the site, so I'm very open to any visual nitpicks or subjective suggestions that you have!

### DIFF
--- a/src/main/webapp/query-control.js
+++ b/src/main/webapp/query-control.js
@@ -30,6 +30,7 @@ function clearPreviousResult() {
   document.getElementById('census-link').style.display = 'none';
   am4core.disposeAllCharts();
   document.getElementById('more-info').innerText = 'Please wait. Loading...';
+  document.getElementById('toggle-data-btn').style.display = 'none';
   document.getElementById('result').style.display = 'block';
 }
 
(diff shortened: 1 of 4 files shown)


src/main/webapp/query-control.js
```diff
@@ -25,12 +25,9 @@ function getColor() {
 
 function clearPreviousResult() {
   document.getElementById('data-table').innerHTML = '';
-  document.getElementById('colors').style.display = 'none';
-  document.getElementById('edit-title').style.display = 'none';
   document.getElementById('year-slider').style.display = 'none';
+  document.getElementById('top-buttons').style.display = 'none';
   document.getElementById('census-link').style.display = 'none';
-  document.getElementById('toggle-data-btn').style.display = 'none';
```
kyracq, line 32: This button doesn't seem to hide on new queries

otobin: Maybe I'm looking at an updated version but for me all of the buttons are hidden when the page loads

shulijones: I also can't seem to replicate what you found, Kyra. Could you try a hard refresh, and if it still persists, let me know exactly what you did to see that? Thanks!

kyracq: If I query something there's no census data on the the toggle data button shows: http://localhost:8080/#person-type=under-18&action=moved&location=04&year=2018

PR #64, opened 2020-08-30
Show recommendations

otobin:
Very small bug fix -- the algorithm wasn't always returning 5 recommendations because the for loop was checking against the current size and not the size before additions.

### DIFF
--- a/src/main/java/RecommendationsServlet.java
+++ b/src/main/java/RecommendationsServlet.java
@@ -164,7 +164,7 @@ public class RecommendationsServlet extends HttpServlet {
     // may not generate 4. In this case, populate the remaining recommendations of the 5 with random
     // ones.
     int recommendationSize = recommendationList.size();
-    if (recommendationList.size() < 5) {
+    if (recommendationSize < 5) {
       for (int i = 0; i < 5 - recommendationSize; i++) {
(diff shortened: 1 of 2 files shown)


src/main/java/RecommendationsServlet.java
```diff
@@ -163,8 +163,9 @@ private void updateFrequencyMaps(ArrayList<VisualizationData> userHistory) {
     // When the user has very similar queries in their search history, the recommendation algorithm
     // may not generate 4. In this case, populate the remaining recommendations of the 5 with random
     // ones.
-    if (recommendationList.size() < 4) {
-      for (int i = 0; i < 5 - recommendationList.size(); i++) {
+    int recommendationSize = recommendationList.size();
+    if (recommendationSize < 5) {
+      for (int i = 0; i < 5 - recommendationSize; i++) {
         recommendationList.add(getRandomRecommendation(userId, userHistory));
```
shulijones, line 169: Could this be a while loop (while recommendationList.size() < 5) ?

nsri-google: yes it seems more like a while loop.

otobin: switched to while loop

anicr7 (approved):

